PR #31, opened 2020-10-06
New Unitary Tests

oualib:
- Correcting Unitary Tests
 - Correcting method search and some magic vDataFrame methods
 - Adding new Unitary Tests


verticapy/tests/vDataFrame/test_vDF_combine_join_sort.py
```diff
@@ -198,6 +198,25 @@ def test_vDF_join(self, market_vd):
         # SELECT COUNT(*) FROM cross_join WHERE Name2 IS NULL;
         assert cross_join["Name2"].count() == 63616
 
+        # join directly with a Vertica table
+        not_dried._VERTICAPY_VARIABLES_["cursor"].execute(
```
afard, line 202: Could we use vDataFrame.to_db() instead?

verticapy/tests/vDataFrame/test_vDF_correlation.py
```diff
@@ -35,27 +36,72 @@ def amazon_vd(base):
...
+            show=False,
         )
-        assert result1_1["value"][0] == result1["value"][0]
-        assert result1_1["value"][1] == result1["value"][1]
-        assert result1_1["value"][2] == result1["value"][2]
-        assert result1_1["value"][3] == result1["value"][3]
-        assert result1_1["value"][4] == result1["value"][4]
-        assert result1_1["value"][5] == result1["value"][5]
+
+        assert result4["value"][1] == pytest.approx(0.669, 1e-2)
```
afard, line 101: Imho, threshold of 1e-2 indicates that there is a big change in each run. Is there any randomness involved in acf and pacf?

verticapy/tests/vDataFrame/test_vDF_statistical_tests.py
```diff
@@ -1,33 +0,0 @@
-# (c) Copyright [2018-2020] Micro Focus or one of its affiliates.
```
afard, line 1: Why have you deleted the file?

afard (commented):
In order to practice the process, I won't approve the PR until you reply my comments.

verticapy/tests/vDataFrame/test_vDF_combine_join_sort.py
oualib: We could add a method for "local temporary tables in 'to_db', it can be a good idea.

verticapy/tests/vDataFrame/test_vDF_correlation.py
oualib: There is no randomness but if multiple dates are similar, the order can change and so it can change the output.

verticapy/tests/vDataFrame/test_vDF_combine_join_sort.py
afard: That might unnecessary for now.

verticapy/tests/vDataFrame/test_vDF_statistical_tests.py
oualib: I changed it to test_vDF_tsa_tools as the stat tests were moved from the vDataFrame.

PR #36, opened 2020-10-10
Testsunit1

oualib:


oualib:
Multiple Corrections + New UT
 - Correcting vCharts bug
 - Erasing the print_info parameter, it is now possible to use the set_display_parameters method of the vdf to choose either to print or not to print the information.
 - Erasing properly the temp relations
 - Sorting correctly the vDataFrame
 - Adding multiple tests for the plot test

PR #47, opened 2020-10-19
Testunits 8

afard:
I have implemented test_linear_regression.py except its get_plot test.
I have also added test_model_from_vDF to test training a model from a vDF, but marked as skip until the feature is implemented.


oualib (approved):
Excellent, I think you can cover easily ElasticNet, Ridge and Lasso which are exactly the same with different parameters.

LinearSVR is also easy to cover because it is very similar.

PR #48, opened 2020-10-22
Testunits 8

afard:
Implementing the tests of test_logistic_regression.py except test_plot and test_set_cursor.


oualib (approved):
Excellent, we can move to the next tests.

PR #50, opened 2020-11-17
Testunits 9

afard:
Added test_elastic_net.py, test_lasso.py, and test_rigid.py


oualib (approved):
I will push the corrections from my side

PR #54, opened 2020-12-03
Preparing 0.3.2

oualib:
- Adding random state
 - Adding stats package
 - simplifying the code

### DIFF
--- a/verticapy/stats.py
+++ b/verticapy/stats.py
@@ -49,7 +49,7 @@
 # Modules
 #
 # Standard Python Modules
-import os, math, shutil, re, time, decimal, warnings
+import os, math, shutil, re, time, decimal, warnings, datetime
 
 # VerticaPy Modules
 import verticapy
@@ -74,6 +74,11 @@ Parameters
 ----------
 expr: object
     Expression.
+
+Returns
+-------
+str_sql
+    SQL expression.
     """
     expr = format_magic(expr)
     return str_sql("ABS({})".format(expr), "float")
@@ -89,6 +94,11 @@ Parameters
 ----------
 expr: object
     Expression.
+
+Returns
+-------
+str_sql
+    SQL expression.
     """
     expr = format_magic(expr)
     return str_sql("ACOS({})".format(expr), "float")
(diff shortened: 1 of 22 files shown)


verticapy/vdataframe.py
```diff
@@ -8783,6 +8790,67 @@ def to_vdf(self, name: str):
         file.close()
         return self
 
+    # ---#
+    def train_test_split(
```
afard, line 8794: It make sense to have the train_test_split function in vDataFrame. Can you please add unit tests for it?

verticapy/tests/stats/test_stats.py
```diff
@@ -0,0 +1,153 @@
...
+    yield titanic
+    with warnings.catch_warnings(record=True) as w:
+        drop_table(
+            name="public.titanic", cursor=base.cursor,
+        )
+
+
+class TestStats:
+    def test_abs(self, titanic_vd):
+        assert st.abs(titanic_vd["age"]) == 'ABS("age")'
```
afard, line 35: I don't understand these tests. Does applying the new functions on a column of vDataFrame return a text?

verticapy/stats.py
```diff
@@ -0,0 +1,591 @@
...
+import os, math, shutil, re, time, decimal, warnings
+
+# VerticaPy Modules
+import verticapy
+import vertica_python
+from verticapy.toolbox import *
+
+#
+# ---#
+pi = str_sql("PI()")
```
afard, line 61: What is `str_sql`?

oualib: It is a new type used to do sql operations allowing the correct SQL code generation.
For example
"2" + "3" = "(2) + (3)" whereas for string it would be 23
abs("age") = "ABS(age)" ...
It is thanks to this type that we can do the amazing syntax.

verticapy/tests/stats/test_stats.py
oualib: it returns an str_sql that when it is compare to str, it will be casted during the comparison.

verticapy/vdataframe.py
oualib: Yes, it is much more easier and it is part of the data preparation. It allows a simple call with much less parameters. As we have this seeded_random, we can now return vDataFrames without creating temporary elements.

afard: Cool!
If you agree, let's treat it as a new feature for vDF. Then, I'll approve the PR after you add tests for it.

afard (approved):

PR #58, opened 2020-12-16
Increasing the version in setup.py from 0.3.2 to 0.4.0

afard:


afard (commented):
It is a very simple change of version in the setup.py before tagging the 0.4.0 release.

PR #65, opened 2021-01-11
Preparing v0.4.1

oualib:
- vDataFrame to_vdf removed and replace by to_pickle
 - utilities read_vdf is removed
 - udf generator
 - durbin_watson returns only d
 - new tests: het_arch, het_breuschpagan, het_goldfeldquandt,  het_white, kurtosistest, normaltest, skewtest, normaltest
 - adding new Unitary Tests and correcting others.
 - new geo functions
 - apply of stats module - new parameters
 - new methods in the regression report: anova/details/metrics
 - new metrics: r2 adjusted Root MSE
 - adding class Pipeline - We need to make UT and correct some functions
 - new function anova_table in metrics
 - correcting the alpha of Linear Models
 - adding the mapping server -> host in the connection creation
 - correcting the setup

### DIFF
--- a/verticapy/learn/pipeline.py
+++ b/verticapy/learn/pipeline.py
@@ -1,0 +1,310 @@
+# (c) Copyright [2018-2020] Micro Focus or one of its affiliates.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# |_     |~) _  _| _  /~\    _ |.
+# |_)\/  |_)(_|(_||   \_/|_|(_|||
+#    /
+#              ____________       ______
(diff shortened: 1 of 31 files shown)


verticapy/vdataframe.py
```diff
@@ -5993,9 +5993,7 @@ def load(self, offset: int = -1):
         """
         check_types([("offset", offset, [int, float],)])
         save = self._VERTICAPY_VARIABLES_["saving"][offset]
-        vdf = {}
-        exec(save, globals(), vdf)
-        vdf = vdf["vdf_save"]
+        vdf = pickle.loads(save)
```
afard, line 5996: I think it would be better if you set the protocol of pickle. The default protocol for Python 3.6 and 3.7 is protocol 3, but it is protocol 4 for Python 3.8.
Protocol version 4 was added in Python 3.4 and is incompatible with previous versions. It adds support for very large objects, pickling more kinds of objects, and some data format optimizations.
I think you should set pickle to use protocol 4 for all supported Python versions.

verticapy/learn/pipeline.py
```diff
@@ -0,0 +1,310 @@
...
+#
+# VerticaPy Modules
+from verticapy import vDataFrame
+from verticapy.utilities import *
+from verticapy.toolbox import *
+from verticapy.errors import *
+from verticapy.learn.vmodel import *
+
+# ---#
+class Pipeline:
```
afard, line 59: Where are the unit tests for Pipeline?

afard (approved):
I approve and merge the PR. Please address my comments in a new PR.

PR #66, opened 2021-01-12
Auto pypi

afard:
This PR adds a workflow which is supposed to publish master to PyPI whenever a tagged commit is pushed to the Git repository.
And it’ll publish any push to TestPyPI which is useful for providing test builds to the alpha users as well as making sure that release pipeline remains healthy.


afard:
It is only an action workflow and I cannot test it without merging. Therefore, I skip the review process.

PR #83, opened 2021-02-08
Preparing v0.5.0

oualib:
What's New in VerticaPy v0.5.0

New features:
Machine Learning

    Algorithms: XGBoostClassifier and XGBoostRegressor are available.
    Additional Parameters for the cross_validate function.
    Additional Metrics AIC & BIC for the regression_report method and the score method. A function aic_bic is also available.
    New functions to evaluate models and to do Hyper Parameters Tuning: grid_search_cv and validation_curve.
    New method for Classifier: Classifier.cutoff_curve.

Soundex

    3 Soundex functions are available to encode or simply compute Levenshtein distance between expressions.

Geospatial

    New Functions: describe_index, intersect, rename_index and drop. They allow the user to manage indexes and to compute the intersection between Polygons and data Points.
    Import / Export SHP files with: read_shp and vDataFrame.to_shp.
    Draw maps with: vDataFrame.geo_plot.

Statistical Tests & Tools

    New function: endogtest to test endogeneity.
    New function: variance_inflation_factor to detect multicollinearity.

vDataFrame

    Removed: DataFrame[].donut
    New method: vDataFrame.recommend allows Collaborative Filtering.
    New method: vDataFrame.polynomial_comb allows easy Multivariate Analysis.

Graphics & Data Visualization

    All Graphics have a new extra parameter: **style_kwds which allows the user to pass any Matplotlib parameters to the function used during the drawing.
    Multiple new methods: vDataFrame.pie, vDataFrame.stacked_area, vDataFrame[].spider.
    Additional Parameters for: vDataFrame.bar [pyramid population bar] and vDataFrame.pie [donut chart & rose chart].

Deprecated & Replaced

    The package verticapy.connections.connect has been moved to verticapy.connect
    The package verticapy.learn.datasets has been moved to verticapy.datasets
    The package verticapy.learn.tsa.tools has been moved to verticapy.stats
    The package verticapy.learn.tsa.models has been moved to verticapy.learn.tsa
    The functions drop_table, drop_text_index, drop_view & drop_model have been replaced by the function: drop

GitHub & Unit Tests

    More unit tests are available.


verticapy/learn/tools.py
```diff
@@ -0,0 +1,611 @@
...
+            cursor.execute(
+                "SELECT * FROM verticapy.models WHERE LOWER(model_name) = '{}'".format(
+                    str_column(name)
+                )
+            )
+            result = cursor.fetchone()
+        except:
+            result = []
+        if result:
+            raise NameError("The model '{}' already exists !".format(name))
```
afard, line 87: The `conn` will be left open if the program raises here. Yon can repeat lines 73 and 74 here.

verticapy/learn/tools.py
```diff
@@ -0,0 +1,611 @@
...
+#
+#
+# Modules
+#
+# VerticaPy Modules
+from verticapy.toolbox import *
+
+#
+# ---#
+def check_model(name: str, cursor=None):
```
afard, line 56: I think it would be much nicer if you rename this function to `does_model_exist`, and return a Boolean. In other words, returning true when the model already exists, and false otherwise.
IMHO, using raised errors instead of conditional statements makes the code less readable and its debugging more difficult. This will lead to more challenging code maintenance and less community contribution.

verticapy/learn/tools.py
```diff
@@ -0,0 +1,611 @@
...
+model
+    The model.
+    """
+    check_types([("name", name, [str],), ("test_relation", test_relation, [str],)])
+    cursor = check_cursor(cursor)[0]
+    try:
+        check_model(name=name, cursor=cursor)
+        raise NameError("The model '{}' doesn't exist.".format(name))
+    except:
+        pass
```
afard, line 119: The code in the `load_model` function is very strange! To check that the model already exits in an expected situation, the `check_model` function is supposed throw an exception. It is very odd to use exceptions for the normal flow of a task. Exception are supposed to be used for stopping the execution of a task, not as a replacement of if-else statements.
It was way more straightforward if the the `check_model` function had returned a true or false value indicating the existence of a model. Then, a `raise` could be used properly for stopping the execution of the task.

verticapy/learn/tools.py
afard: Or, you may make the `check_model` function return an integer: 0 if the model doesn't exist, 1 if it is of type1 (native), and 2 if it's of type2 (non-native).

verticapy/learn/tools.py
```diff
@@ -0,0 +1,611 @@
...
+    if model_type == "svm_classifier":
+        parameters = "".join(info[1].split("class_weights=")[1].split("'"))
+        parameters = parameters[3 : len(parameters)].split(",")
+        del parameters[0]
+        parameters += [
+            "class_weights=" + info[1].split("class_weights=")[1].split("'")[1]
+        ]
+    elif model_type != "svd":
+        parameters = info[1].split(",")
+    if model_type != "svd":
```
afard, line 325: Why the if condition here is similar to the condition of elif on line 323?

verticapy/learn/tools.py
oualib: Thanks for the idea! It is done and it will be pushed after the v0.5.0 release.

verticapy/learn/tools.py
oualib: Thanks for the idea! It is done and it will be pushed after the v0.5.0 release.

verticapy/learn/tools.py
oualib: It is done and it will be pushed after the v0.5.0 release.

verticapy/learn/tools.py
```diff
@@ -0,0 +1,611 @@
...
+                        try:
+                            result_tmp = float(result_tmp)
+                        except:
+                            pass
+                        if result_tmp == None:
+                            result_tmp = "None"
+                        model_save[elem[0]] = result_tmp
+                except:
+                    raise
+                    model_save = {}
```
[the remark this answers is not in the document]
oualib: Corrected, thanks!

verticapy/learn/tools.py
```diff
@@ -0,0 +1,611 @@
...
+    except:
+        try:
+            cursor.execute(
+                "SELECT GET_MODEL_SUMMARY (USING PARAMETERS model_name = '"
+                + name
+                + "')"
+            )
+            info = cursor.fetchone()[0].replace("\n", " ")
+            info = "kmeans(" + info.split("kmeans(")[1]
+        except:
```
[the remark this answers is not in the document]
oualib: good idea!

verticapy/tests/vModel/test_lasso.py
```diff
@@ -147,10 +150,15 @@ def test_to_sklearn(self, model):
         prediction = model.cursor.fetchone()[0]
         assert prediction == pytest.approx(md.predict([[3.0, 11.0, 93.0]])[0][0])
 
-    @pytest.mark.skip(reason="shap doesn't want to work on python3.6")
-    def test_shapExplainer(self, model):
-        explainer = model.shapExplainer()
-        assert explainer.expected_value[0] == pytest.approx(5.81837771)
+    try:
+        import shap
```
afard, line 154: I think we skipped this test because adding the `shap` package to the deps list in tox.ini was causing failure under Py3.6 in the previous test system. Now that it is successfully added, you should be able to run this test normally without any limit.
Btw, you can skip conditionally using the `skipif` marker:
https://docs.pytest.org/en/latest/skipping.html#skipif

verticapy/tests/vModel/test_lasso.py
```diff
@@ -147,10 +150,15 @@ def test_to_sklearn(self, model):
...
+
+        def test_shapExplainer(self, model):
+            explainer = model.shapExplainer()
+            assert explainer.expected_value[0] == pytest.approx(5.81837771)
+
+    except:
+        pass
```
afard, line 161: If you prefer not to use the `skipif` marker for any reason, you'd better to call the `pytest.skip(reason)` function here instead of `pass`.

verticapy/tests/vModel/test_elastic_net.py
```diff
@@ -148,10 +151,15 @@ def test_to_sklearn(self, model):
...
-        assert explainer.expected_value[0] == pytest.approx(5.81837771)
+    try:
+        import shap
+
+        def test_shapExplainer(self, model):
+            explainer = model.shapExplainer()
+            assert explainer.expected_value[0] == pytest.approx(5.81837771)
+
+    except:
+        pass
```
afard, line 162: The same here. Please use `pytest.skip(reason)` instead of `pass`. The reason can be something like "unsupported configuration".

verticapy/learn/tools.py
afard: OK

verticapy/tests/vModel/test_lasso.py
afard, line 161: It is important to fix this; otherwise, how can we know if the test is really performed.
Please use `pytest.skip(reason)` instead of `pass`.

verticapy/tests/vModel/test_linear_regression.py
```diff
@@ -145,10 +148,15 @@ def test_to_sklearn(self, model):
...
-        assert explainer.expected_value[0] == pytest.approx(5.81837771)
+    try:
+        import shap
+
+        def test_shapExplainer(self, model):
+            explainer = model.shapExplainer()
+            assert explainer.expected_value[0] == pytest.approx(5.81837771)
+
+    except:
+        pass
```
afard, line 159: Please fix this test. It should either fail or conditionally skipped.

verticapy/tests/vModel/test_linear_svc.py
```diff
@@ -130,12 +143,15 @@ def test_to_sklearn(self, model):
...
-    @pytest.mark.skip(reason="shap doesn't want to work on python3.6")
-    def test_shapExplainer(self, model):
-        explainer = model.shapExplainer()
-        assert explainer.expected_value[0] == pytest.approx(-0.22667938806360247)
+        def test_shapExplainer(self, model):
+            explainer = model.shapExplainer()
+            assert explainer.expected_value[0] == pytest.approx(-0.22667938806360247)
+
+    except:
+        pass
```
afard, line 154: Please fix this test. It should either fail or conditionally skipped.

verticapy/tests/vModel/test_linear_svr.py
```diff
@@ -146,10 +149,15 @@ def test_to_sklearn(self, model):
...
-        assert explainer.expected_value[0] == pytest.approx(5.819113657580594)
+    try:
+        import shap
+
+        def test_shapExplainer(self, model):
+            explainer = model.shapExplainer()
+            assert explainer.expected_value[0] == pytest.approx(5.819113657580594)
+
+    except:
+        pass
```
afard, line 160: Please fix this test. It should either fail or be conditionally skipped.

verticapy/tests/vModel/test_logistic_regression.py
```diff
@@ -140,10 +153,15 @@ def test_to_sklearn(self, model):
...
-        assert explainer.expected_value[0] == pytest.approx(-0.4617437138350809)
+    try:
+        import shap
+
+        def test_shapExplainer(self, model):
+            explainer = model.shapExplainer()
+            assert explainer.expected_value[0] == pytest.approx(-0.4617437138350809)
+
+    except:
+        pass
```
afard, line 164: Please fix this test. It should either fail or be conditionally skipped.

verticapy/tests/vModel/test_ridge.py
```diff
@@ -154,10 +158,15 @@ def test_to_sklearn(self, model):
...
-        assert explainer.expected_value[0] == pytest.approx(5.81837771)
+    try:
+        import shap
+
+        def test_shapExplainer(self, model):
+            explainer = model.shapExplainer()
+            assert explainer.expected_value[0] == pytest.approx(5.81837771)
+
+    except:
+        pass
```
afard, line 169: Please fix this test. It should either fail or be conditionally skipped.

verticapy/tests/vModel/test_xgb_classifier.py
```diff
@@ -0,0 +1,419 @@
...
+
+            @pytest.mark.skip(reason="not yet available.")
+            def test_shapExplainer(self, model):
+                explainer = model.shapExplainer()
+                assert explainer.expected_value[0] == pytest.approx(
+                    -0.22667938806360247
+                )
+
+        except:
+            pass
```
afard, line 206: Please fix this test. It should either fail or be conditionally skipped.

verticapy/tests/vModel/test_xgb_classifier.py
```diff
@@ -0,0 +1,419 @@
...
+            )
+            prediction = model.cursor.fetchone()[0]
+            assert prediction == pytest.approx(
+                md.predict([["Bus", "Male", 0, "Cheap", "Low"]])[0]
+            )
+
+        try:
+            import shap
+
+            @pytest.mark.skip(reason="not yet available.")
```
afard, line 198: The test is skipped even when `shap` is successfully imported :)

verticapy/tests/utilities/test_utilities.py
```diff
@@ -12,49 +12,276 @@
...
+        base.cursor.execute(
+            "CREATE TEXT INDEX drop_index ON drop_data (id, transportation);"
+        )
+        drop("drop_index", base.cursor)
+        try:
+            base.cursor.execute("SELECT * FROM drop_index;")
+            fail = False
+        except:
+            fail = True
+        assert fail
```
afard, line 134: Pytest has a standard way of asserting exceptions:
https://docs.pytest.org/en/reorganize-docs/new-docs/user/pytest_raises.html
https://docs.pytest.org/en/stable/reference.html#pytest-raises

verticapy/tests/utilities/test_utilities.py
```diff
@@ -12,49 +12,276 @@
...
+            "world_polygons", base.cursor,
+        )
+        try:
+            describe_index(
+                "world_polygons", base.cursor, True,
+            )
+            fail = False
+        except:
+            fail = True
+        assert fail
```
afard, line 173: It is better the pytest.raises for asserting exceptions.

verticapy/learn/tsa/__init__.py
```diff
@@ -44,3 +44,1801 @@
...
+from verticapy.learn.linear_model import LinearRegression
+from verticapy import vDataFrame
+from verticapy.plot import gen_colors
+from verticapy.learn.tools import *
+
+# Other Python Modules
+from dateutil.parser import parse
+import matplotlib.pyplot as plt
+
+# ---#
```
afard, line 65: Why have you moved the implementation of classes and functions into the __init__.py file?
Afaik, the primary use of __init__.py is to initialize Python packages and not to hold content.

afard: I think it would be more appropriate if you move them into a base.py file in this directory, and import it here if needed.

verticapy/learn/tsa/__init__.py
```diff
@@ -44,3 +44,1801 @@
...
+        )
+        result = vdf_from_relation(final_relation, "VAR", self.cursor,)
+        if nlead > 0:
+            for elem in X:
+                result[elem].apply(
+                    "CASE WHEN {} >= '{}' THEN NULL ELSE {} END".format(
+                        ts, first_t, "{}"
+                    )
+                )
+        return result
```
afard, line 1844: Do we have test coverage for all TSA functionalities?

oualib: Not yet

verticapy/learn/tsa/__init__.py
oualib: Because I'll move it to tsa.py just right after when I will erase the deprecated message which is in the files models and tools.

verticapy/tests/base.py
```diff
@@ -208,3 +209,27 @@ def require_DB_version_at_least(self, min_version):
...
+base_test.setUp()
+
+if not (
+    os.path.isfile(os.path.dirname(verticapy.__file__) + "/tests/verticaPy_test.conf")
+):
+    path = os.path.dirname(verticapy.__file__) + "/tests/verticaPy_test.conf"
+    confparser = ConfigParser()
+    confparser.optionxform = str
+    try:
+        confparser.read(path)
```
afard, line 224: I don't see what is the point of this line.

verticapy/tests/base.py
```diff
@@ -208,3 +209,27 @@ def require_DB_version_at_least(self, min_version):
...
+        pass
+    if confparser.has_section("vp_test_config"):
+        confparser.remove_section("vp_test_config")
+    confparser.add_section("vp_test_config")
+    for elem in base_test.test_config:
+        if elem != "log_level":
+            confparser.set("vp_test_config", elem, str(base_test.test_config[elem]))
+    f = open(path, "w+")
+    confparser.write(f)
+    f.close()
```
afard, line 235: Apparently, you don't clean a potentially created file, after the tests are done.
I think it would be nicer if you move the code under the `if not` into a function defined in conftest.py. Also, define another function over there that deletes the file. Then, you can call the create_file function before the `yield`, and the delete_file function after  the `yield` in the conftest.base() function.
Also, I suggest you name the file different from vp_test_config, so we won't over write it if it already exists.

verticapy/learn/tools.py
oualib: Done;

verticapy/learn/tools.py
oualib: Done;

verticapy/learn/tools.py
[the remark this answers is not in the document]
oualib: Done;

verticapy/learn/tools.py
oualib: Corrected;

verticapy/tests/vModel/test_lasso.py
oualib: OK

verticapy/tests/utilities/test_utilities.py
oualib: Corrected

verticapy/tests/utilities/test_utilities.py
oualib: corrected

verticapy/tests/base.py
oualib: Corrected

verticapy/tests/base.py
oualib: Done;

afard: When do you push your fixes to the PR?

verticapy/tests/base.py
```diff
@@ -20,6 +20,7 @@
 import inspect
 import getpass
 import vertica_python
+import verticapy
```
afard, line 23: I think this import is redundant.

afard (approved):

PR #108, opened 2021-06-02
new version 0.6.2

oualib:
- Handling temporary local tables for read_csv, pandas_to_vertica, load_dataset...
 - AutoML works with previous Vertica versions
 - plot_pca_circle is now available
 - correction of some bugs (load_model, pie, sql, fillna, asfreq, filter...)
 - New functions: cochrane_orcutt, balance
 - param changes: pandas_to_vertica, read_csv, read_json


verticapy/datasets.py
```diff
@@ -226,9 +226,16 @@ def load_dataset(
...
+            schema = "v_temp_schema"
+            temp = "LOCAL TEMPORARY "
+        else:
+            schema = str_column(schema)
+            temp = ""
         cursor.execute(
-            "CREATE TABLE {}.{}({});".format(
-                str_column(schema), str_column(name), str_create,
+            "CREATE {}TABLE {}({});".format(
+                temp, str_column(schema) + "." + str_column(name) if schema != 'v_temp_schema' else str_column(name), str_create,
```
afard, line 238: (1) You should add `ON COMMIT PRESERVE` when you create a temp table; otherwise, its data will be removed after a commit. Look at this page for more info: https://www.vertica.com/docs/10.1.x/HTML/Content/Authoring/SQLReferenceManual/Statements/CREATETEMPORARYTABLE.htm
(2) It would be nice to have vDFs pointing to temp tables, but please be advised that ML training functions currently don't accept temp tables as their input. Therefore, it makes sense if you postpone this change until it is supported on Vertica ML side as well. Otherwise, we'll end up with vDFs that don't work properly for some VerticaPy functions.

verticapy/vdataframe.py
```diff
@@ -3256,6 +3256,59 @@ def bar(
...
+        last_count, last_elem, n = topk["count"][-1], topk["index"][-1], len(topk["index"])
+        if method == "over":
+            last_count = last_count * x
+        elif method == "under":
+            last_count = last_count / x
+        vdf = self.search("{} = '{}'".format(column, last_elem))
+        for i in range(n - 1):
+            vdf = vdf.append(self.search("{} = '{}'".format(column, topk["index"][i])).sample(n = last_count))
+        vdf.sort(order_by)
+        return vdf
```
afard, line 3310: Why don't you use the in-DB BALANCE function?
 https://www.vertica.com/docs/10.1.x/HTML/Content/Authoring/SQLReferenceManual/Functions/MachineLearning/BALANCE.htm

afard:
@oualib,
I went through the PR once. My main concern is that vDFs based on temp table will fail on some ML functions because meta-functions of Vertica-ML do not accept temp tables at this time.

verticapy/vdataframe.py
oualib: Because it creates a view. They can still use it in the learn.preprocessing package!

verticapy/datasets.py
oualib: I just corrected it. It will be available in the next PR

oualib: Agree!

afard (approved):

PR #184, opened 2021-11-02
Update dataset example

MiConnell:
Update the dataset example to reference `verticapy.datasets` instead of `verticapy.learn.datasets` which throws a `ModuleNotFoundError`


oualib (approved):
Thanks for noticing this non-updated line!

PR #193, opened 2021-12-02
Fixing tests

afard:
In test_plot_tree for different model the following assert causes problem because of an extra '\n' at the end of result.source.
`assert model.to_graphviz() == result.source`


verticapy/tests/vDataFrame/test_vDF_plot.py
```diff
@@ -228,6 +228,7 @@ def test_vDF_bar(self, titanic_vd, amazon_vd):
         )
         plt.close("all")
 
+    @pytest.mark.skipif(sys.version_info >= (3, 7), reason="this test is incompatible with newer versions of matplotlib")
```
afard, line 231: The matplotlib installed in the test environment for Python 3.7 and 3.8 is 3.5.0 which was released on November 15, 2021. The way several tests in this file are written is incompatible with this new version of the package. The environment installs matplotlib 3.3.4 for Python 3.6, so these tests are still passing for that version.
We should improve these tests not to rely on internal implementation of the matplotlib package.

verticapy/tests/vDataFrame/test_vDF_plot.py
```diff
@@ -293,12 +295,10 @@ def test_vDF_bubble(self, iris_vd, titanic_vd,):
         assert max([elem[1] for elem in result3.get_offsets().data]) <= 7.9
         plt.close("all")
 
+    @pytest.mark.skipif(sys.version_info >= (3, 7), reason="this test is incompatible with newer versions of matplotlib")
     def test_vDF_density(self, iris_vd):
         # testing vDataFrame[].density
-        try:
-            create_verticapy_schema(iris_vd._VERTICAPY_VARIABLES_["cursor"])
-        except:
-            pass
```
afard, line 301: Passing an exception blinds us on any potential problem. We should try to avoid it.

oualib (approved):

Issue #265, opened 2022-02-25
Bug with pandas_to_vertica function - backslash escape character


gaetan-dion:
When we use pandas_to_vertica, some files, pandas DataFrame or vDataframe have specials characters inside, like \ (backslash).   
We observe that these lines are not injected in a Vertica table.   

The cause is here : [datasets.py#L245](https://github.com/vertica/VerticaPy/blob/a188213576fa71e5e1623d1cb46fd4f14cc8e41f/verticapy/datasets.py#L245)

The backslash is used like escape character, but if there are some backslash inside the file also, the concerned line(s) can't be injected.   
I advise to nest the copy request between 3 double quotes : """copy ... """

oualib:
Hi Gaetan,

Thanks for reporting the issue.
I changed the default parameter of escape '\\' to '\010' which is the ascii code of \n. In the next PR, this function should work perfectly fine.

Badr

oualib:
Hi Gaetan,

Thanks for reporting the issue.
I changed the default parameter escape in the read_csv function: '\\' to '\027' which is the ascii code of ESC. In the next PR, this function should work perfectly fine.

Badr

Issue #262, opened 2022-02-21
KNeighborsRegressor, KNeighborsClassifier


lepelletieralexandre:
Hello,

for KNeighborsRegressor, KNeighborsClassifier model, the method get_attr doesn't work. 

![image](https://user-images.githubusercontent.com/49387599/154975886-c4f7bf57-9479-412d-9725-dd6303b4fd02.png)

Regards,

Alexandre

Issue #260, opened 2022-02-21
pandas_to_vertica - possibility to change the csv path


gaetan-dion:
https://github.com/vertica/VerticaPy/blob/master/verticapy/utilities.py#L503   

In this line, pandas_to_vertica function write a temporary csv file, before sending the data to Vertica.   
We need to can choose the path of this file, because sometimes, we don't have access right to the default path.

oualib:
Hi Gaetan,

A new parameter 'temp_path' will be added to the function in the next PR.

Badr

Issue #255, opened 2022-02-11
sample, method systematic and stratified


lepelletieralexandre:
Hello,

There is a problem with sample function for systematic and stratified method. 

In fact, where 1/x is far from an integer (for exemple 3.4) the ratio is far. For exemple, titanic dataset (1234 lines) and x = 0.29, we want 358 lines, but we have 411 lines. And for x = 0.33 lines, we want 408 lines but we have 411 lines (same as x = 0.29). 

with this exemple, the solution are more stable : 

```
print("ratio : ", 1/x)
titanic.eval("id", "row_number() over()")
titanic["id"].apply("id")
titanic.eval("id2", "cast(id*" + str(x) + " as integer)")
titanic.cummin("id", "id2", name = 'id3')
print("true value : ",len(titanic)*x)
print("test value : ", len(titanic.search("id3 = id")))
```

here are the results we have for some values of x : 

x = 0.29 : 

![image](https://user-images.githubusercontent.com/49387599/153619113-2240df5d-cf14-4c61-9a8f-9434dcbcbf6f.png)

x = 0.33 : 

![image](https://user-images.githubusercontent.com/49387599/153619164-5e333eb4-2c1b-4178-aa66-dc93032e20b6.png)

x = 0.2 : 

![image](https://user-images.githubusercontent.com/49387599/153619206-2437c2ee-9f3e-4b8c-8fcf-df832f0f2090.png)

x = 0.22 : 

![image](https://user-images.githubusercontent.com/49387599/153619272-47e8e69a-d767-4589-b13b-e74907f19665.png)

your solution is better only when 1/x = integer, but your solution is far when x is far from an interger. 

Can you improve that? 

Regards,

Alexandre

PR #266, opened 2022-02-26
Correcting multiple issues (bug + improvements)

oualib:
- pandas_to_vertica
Solves #265 #260 #253 #197
 - Flex Tables name
Solves #186
 - public schema issues
Solves #209
 - precision error
Solves #216
 - predict & get_attr method of KNN
Solves #223 #262
 - sampling
Solves #254 #255

### DIFF
--- a/verticapy/tests/utilities/test_utilities.py
+++ b/verticapy/tests/utilities/test_utilities.py
@@ -359,6 +359,10 @@ class TestUtilities:
         )
         assert result["verticapy test *+"].avg() == 1.0
 
+    # TODO
+    def test_set_option(self, base):
+        pass
+
     def test_version(self, base):
         result = version(base.cursor,)
         assert result[0] < 20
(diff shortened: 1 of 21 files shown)


verticapy/tests/utilities/test_utilities.py
```diff
@@ -359,6 +359,10 @@ def test_vdf_from_relation(self, base):
         )
         assert result["verticapy test *+"].avg() == 1.0
 
+    # TODO
```
afard, line 362: Instead of having a fake passing test, it would be better to mark it with `skip` decorator. Then, it would be quite easy to monitor which tests are skipped and why.
You may replace `# TODO` with `@pytest.mark.skip(reason="this test will be implemented later")`

afard (approved):

ansleis:
Looks good, made some small edits.

Also, to_parquet needs the fileMode and dirMode parameters to be of type str (instead of int).

Issue #326, opened 2022-05-15
Impossible to cancel randomness seeding


oualib:
When seeding the randomness using the 'set_option' function, it is impossible to set it to None again.

PR #329, opened 2022-05-16
Correcting many bugs

oualib:
Solves #327, #328, #326, #325
 - I named the version 0.9.0-0 to avoid collisions when pulling it. The final one will be 0.9.1
 - Correcting the different bugs listed above.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ with open("README.md", "r") as fh:
 
 setuptools.setup(
     name="verticapy",
-    version="0.9.0",
+    version="0.9.0-0",
     author="Badr Ouali",
     author_email="badr.ouali@vertica.com",
     url="https://github.com/vertica/VerticaPy",
--- a/verticapy/__init__.py
+++ b/verticapy/__init__.py
@@ -46,7 +46,7 @@
 # data around for processing, VerticaPy brings the logic to the data.
 #
 #
-__version__ = "0.9.0"
+__version__ = "0.9.0-0"
 __author__ = "Badr Ouali"
 __author_email__ = "badr.ouali@vertica.com"
 __description__ = (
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -19,7 +19,7 @@
 
 setuptools.setup(
     name="verticapy",
-    version="0.9.0",
+    version="0.9.0-0",
```
afard, line 22: Let's keep the version numbers simple and only 3 digits: so `0.9.1` instead of `0.9.0-0`.

verticapy/__init__.py
```diff
@@ -46,7 +46,7 @@
 # data around for processing, VerticaPy brings the logic to the data.
 #
 #
-__version__ = "0.9.0"
+__version__ = "0.9.0-0"
```
afard, line 49: I think it was decided in the past that version of VertiaPy has 3 parts: MAJOR.MINOR.PATCH. So, it should be `0.9.1` instead of `0.9.0-0`.

setup.py
oualib: I know but we will not release it. It is mainly to avoid collision when a user is pulling the repo with the version 0.9.1 (not the real one but just the last merge), he will not be able to upgrade to 0.9.1 (when this version is released). This is why I am adding an extra digit.

verticapy/__init__.py
oualib: I know but we will not release it. It is mainly to avoid collision when a user is pulling the repo with the version 0.9.1 (not the real one but just the last merge), he will not be able to upgrade to 0.9.1 (when this version is released). This is why I am adding an extra digit.

setup.py
afard: Did we have such a problem before? Are you planning to increase the forth digit on every PR from now on?

oualib: Yes we had. No, I am adding this digit only for the PR which are not releases

afard (approved):

PR #353, opened 2022-07-26
Integrate JQuery datatables into vdataFrame tables (#345)

roypaulin:
This is 1st PR of a series whose goal will be to support interactive tables i.e dynamic tables using javascript.

For now you can test it with:

    init_interactive(all_interactive=True): this will enable interactive table on all vdataframes.
    vdf.idisplay(): use this method if you only one to activate interactive table on the current vdataframe.

This is achieved using the [jQuery dataTables library](https://datatables.net/). It comes with its own style, so although we can make it as similar as possible as a normal vdataframe html table there will be some differences. For now there are only the search bar, the page length and the previous/next buttons are supported but more features will be added.

The table style will be changed progressively in order to have a very dynamic one.

It has been tested with quite a few datasets but feel free to test it on your own datasets.
An interactive table will look like the picture below:
image

### DIFF
--- a/verticapy/__init__.py
+++ b/verticapy/__init__.py
@@ -79,6 +79,7 @@ verticapy.options = {
     "colors": [],
     "color_style": "default",
     "connection": {"conn": None, "section": None, "dsn": None},
+    "interactive": False,
     "count_on": False,
     "footer_on": True,
     "max_columns": 50,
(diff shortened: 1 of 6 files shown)


oualib (approved):
AMAZING WORK Roy! We need to add this option to the next documentation!

PR #358, opened 2022-08-16
changing version id

oualib:


setup.py
```diff
@@ -19,7 +19,7 @@
 
 setuptools.setup(
     name="verticapy",
-    version="0.9.1",
+    version="0.10.0",
```
PriyankaMF, line 22: Considering pervious release had incompatible changes and this one introduces new functionality, shall we call this 1.1.0?

See https://semver.org/

afard (approved):

setup.py
afard: We had decided to keep the main version (the first digit) 0 to indicate that it is still a beta version. It is not unusual for open-source projects to stay in the beta phase for years while they become mature and robust. As a matter of fact, they can become quite popular and be adapted by many customers while they are in the beta phase.
It took about 6 years for our popular Python driver (https://github.com/vertica/vertica-python) to reach to its 1.0.0 version. Of course, I hope that this period will be shorter for VerticaPy. :)

PriyankaMF: Understood. Thanks. As discussed, for minor/patch, let's follow semantic versioning.

PR #366, opened 2022-09-15
Fix connection backup_server_node parameter

lassiraa:
This fixes a SyntaxError with the backup_server_node setting. The code attempts to execute: id_port = 11.111.111.1 when the desired functionality is id_port = '11.111.111.1'.


oualib (approved):

oualib:
Thanks a lot @lassiraa for fixing this issue!

Issue #368, opened 2022-09-28
PC Algorithm


zacandcheese:
Causal inference is a powerful idea and is currently being used to predict the stock market, explain models, and discover the root of many problems. The PC Algorithm is a simple implementation that is effective at producing those results.

PR #369, opened 2022-09-28
PC Algorithm (#368)

zacandcheese:


oualib:
@zacandcheese it seems the tests are not passing. Can you please check why?
@afard Can you please review the code?
@ansleis Can you please review the description of the new features?

afard:
> 

I will try to do it by the end of the week.

afard:
Sorry, I closed the PR by mistake!

oualib:
@afard any news on this one?

afard:
> 

Unfortunately, I was busy last week. Hopefully, I'll review it this week.

afard:
@oualib, I think the tests for Python 3.8 fail because the default version of matplotlib on 3.8 has changed from 3.5.3 to 3.6.1. Apparently, many of our tests rely on internal structures of matplotlib that can be frequently changed between their versions. Can you please look at the failed tests and improve them?

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+# ---#
+def is_dir_path(adj_list: dict, X:str, Y:str):
+    """
+---------------------------------------------------------------------------
+Determines if there is a directed path from X -> Y given an adjacency list.
+
+Parameters
+----------
+adj_list: dict
+    Adjacency List where the values are the other keys
```
afard, line 1438: It is nice that this function is added as a generic tool that might be used in the future somewhere out of the scope of this PR. Nevertheless, it would be nice for the same reason to make its functionality as clear as possible.
For example, is it a correct assumption that the `adj_list` represents a generic directed graph (with potential cycles)?

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+    if Y not in adj_list.keys():
+        raise ParameterError(f"{Y} not in Adjacency List")
+
+    visited = [X]
+    queue = [X]
+
+    while(queue):
+        m = queue.pop(0)
+        for neighbor in adj_list[m]:
+            if m not in adj_list[neighbor]:
```
afard, line 1460: I don't understand this if-statement!
Let's your graph as only X and Y nodes with a cycle:
`adj_list = {{X, {Y}}, {Y, {X}}}`
There is a path from X to Y and a path from Y to X, but the implemented function will deny existence of these paths. Do I miss something?

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
         raise ParameterError(
             f"Wrong type for Parameter X_idx.\nExpected integer, found {type(X_idx)}."
         )
+
+# ---#
+def is_dir_path(adj_list: dict, X:str, Y:str):
```
afard, line 1430: Please add an independent unit test for this function.

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+                    return True
+                if neighbor not in visited:
+                    visited.append(neighbor)
+                    queue.append(neighbor)
+    return False
+    
+
+
+# ---#
+def conditional_chi_square(vdf: vDataFrame, X: str, Y: str, Z:list = [], alpha:float = None):
```
afard, line 1471: Please add an independent unit test for this function. I think you can add your test to https://github.com/vertica/VerticaPy/tree/master/verticapy/tests/stats/test_stats.py

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+    Input vColumn to Test.
+Y: str
+    Input vColumn to Test.
+    .
+Z: list
+    Conditional vColumns
+alpha: float
+    The probability the given distribution was generated randomly
+    Smaller alpha (ie close to 0) lead to denser graphs
+    Larger  alpha (ie close to 1) lead to sparser graphs
```
afard, line 1491: It seems that the valid range of alpha is (0,1). It would be nice if you clarify that in the comments. It would be also very nice if you validate the input alpha value.

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+----------
+vdf: vDataFrame
+    Input vDataFrame.
+X: str
+    Input vColumn to Test.
+Y: str
+    Input vColumn to Test.
+    .
+Z: list
+    Conditional vColumns
```
afard, line 1487: Apparently, you assume that people know what are X, Y, Z and that is why their introducing comments are vague.
The function is supposed to determine the conditional correlation between X and Y given Z as the condition, right? IMHO, it would be very nice if you clarify that for any future contributors.

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+    if isinstance(vdf, str):
+        vdf = vDataFrameSQL(relation=vdf)
+    
+    vdf.are_namecols_in(X) 
+    vdf.are_namecols_in(Y)   
+    vdf.are_namecols_in(Z)
+
+    relation = vdf.__genSQL__()
+    if Z == []:
+        table_sql = "SELECT {0}, {1}, count(*) as nij FROM {2} GROUP BY {0}, {1}".format(X, Y, relation)
```
afard, line 1519: Do you assume that X and Y are categorical columns (like char or varchar )? What if they are numeric (like int or float)?

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+        contingency_table = vDataFrame(sql=table_sql)
```
afard, line 1520: It would be much more elegant if you use the groupby function of vDataFrame directly on `vdf`.
https://www.vertica.com/python/documentation_last/vdataframe/main-methods/groupby/index.php

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+        crosstab_panda = contingency_table.to_pandas()
```
afard, line 1521: The main philosophy of VerticaPy is processing data while it is inside a Vertica database. Therefore, we should try to refrain from extracting data out of the database to the Python environment.

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+        pivot = crosstab_panda.pivot(index=X, columns = Y).fillna(0)
```
afard, line 1522: Do you know that vDataFrame has also a pivot function?
https://www.vertica.com/python/documentation_last/vdataframe/main-methods/pivot/index.php

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+        chi, _, dof, _ = chi2_contingency(pivot)
```
afard, line 1523: Couldn't use the pivot_table_chi2 function of vDataFrame?
https://www.vertica.com/python/documentation_last/vdataframe/main-methods/pivot_table_chi2/index.php

verticapy/stats/tools.py
```diff
@@ -1424,3 +1425,124 @@ def variance_inflation_factor(vdf: vDataFrame, X: list, X_idx: int = None):
...
+        pval = 1 - chi2.cdf(chi, df=dof)
+        if alpha == None:
+            return (chi, pval, dof)
+        return pval > alpha
+    
+    chi = 0
+    dof = 0
+    table_sql = "SELECT {0}, {1}, {2}, count(*) as nij FROM {3} GROUP BY {0}, {1}, {2}".format(', '.join(Z), X, Y, relation)
+    contingency_table = vDataFrame(sql=table_sql)
+    crosstab_panda = contingency_table.to_pandas()
```
afard, line 1533: Same as above!

verticapy/learn/causal_model.py
```diff
@@ -0,0 +1,343 @@
...
+ 
+	# ---#
+	def __init__(self, input_relation : Union[str, vDataFrame]):
+		check_types([("input_relation", input_relation, [str, vDataFrame])])
+		vdf = input_relation
+		if isinstance(vdf, str):
+			vdf = vDataFrameSQL(relation=vdf)
+
+		self.input_relation = vdf
+		self.vdf = vdf.copy()
```
afard, line 37: Why do we need two copies? Perhaps, you can add a short comment to briefly explain what are `self.input_relation` and `self.vdf` and why both should be available.

verticapy/learn/causal_model.py
```diff
@@ -0,0 +1,343 @@
...
+			remove_cols = set(temp._VERTICAPY_VARIABLES_['columns']) - set(input_columns)
+			temp.drop(remove_cols)
+
+		if exclude_columns != []:
+			remove_cols = set(exclude_columns) - set(temp._VERTICAPY_VARIABLES_['columns'])	
+			temp.drop(remove_cols)
+
+		temp.dropna()
+		for col in temp._VERTICAPY_VARIABLES_['columns']:
+			if temp[col].isnum():
```
afard, line 101: Is `max_unique` only applied to numerical columns? What about the categorical columns? Can they have number of categories bigger than `max_unique`?

verticapy/learn/causal_model.py
```diff
@@ -0,0 +1,343 @@
...
+		cit = conditional_chi_square # Conditional Independence Test
+		show_progress = show_progress and vp.options["tqdm"] # Verticapy Override option
+		if show_progress:
+			progress_bar = tqdm(total=max_cond_vars)
+			progress_bar.set_description("Using n conditional variables: 0")
+
+		# Discover the skeleton
+		for cond_set_size in range(max_cond_vars):
+			if method == 'stable':
+				neighbors = {x: skeleton_edges[x].copy() for x in columns}
```
afard, line 214: Except this line, every other line is the same for `method == 'stable'` and `method == 'orig'`; therefor you can make an if-statement only for this line and consolidate the rest. Btw, you can make sure `method` has a valid value before the for-loop.

afard: Ah, I see that there is another difference between the usage of `neighbors` and `skeleton_edges`. So you can copy the right one to a temp variable to consolidate the code, but it would be worth it if they are always small. I'll leave the judgment to you. That said, I think it would still be cleaner if you validate `method` before the for-loop.

afard:
The failing tests on Python 3.8 have been addressed on master. Therefore, updating this branch with master will make the tests pass.

verticapy/learn/causal_model.py
```diff
@@ -0,0 +1,343 @@
...
+					if z not in sepsets[frozenset((x,y))]:
+						if x in pdag[z]:
+							pdag[z].remove(x)
+						if y in pdag[z]:
+							pdag[z].remove(y)
+		
+		edges_to_be_oriented = True
+		while edges_to_be_oriented:
+			remaining_edges = sum([len(pdag[x]) for x in columns])
+		# 2 Orientate non-colliders X -> Z - Y into X -> X -> Y
```
afard, line 291: Did you mean `X -> Z -> Y`?

verticapy/learn/causal_model.py
```diff
@@ -0,0 +1,343 @@
...
+			for (x,z) in node_pairs:
+				if z in pdag[x] and x not in pdag[z]:
+					for y in pdag[z]:
+						if z in pdag[y] and y not in skeleton_edges[x]:
```
afard, line 295: I don't understand the reason of the `y not in skeleton_edges[x]` condition!

verticapy/learn/causal_model.py
```diff
@@ -0,0 +1,343 @@
...
+				if x in pdag[y] and y in pdag[x]:
+					if (is_dir_path(pdag, x, y)):
+						pdag[y].remove(x)   
+		# 4 For each X - Z - Y with X -> W, Y -> W and Z-W orienate Z->W
+			for (z,w) in node_pairs:
+				if w in pdag[z] and z in pdag[w]:
+					for (x,y) in itertools.combinations(pdag[z], 2):
+						if z in pdag[x] and z in pdag[y] and w in pdag[x] and w in pdag[y] and x not in pdag[w] and y not in pdag[w]:
+							pdag[w].remove(z)
+			edges_to_be_oriented = remaining_edges > sum([len(pdag[x]) for x in columns])
```
afard, line 308: I don't understand the logic of this line. Can you please briefly explain it in a short comment?

verticapy/stats/tools.py
zacandcheese: It is correct to assume it represents a generic directed graph. The values of the dictionary can either be a set or list. While I do think this could work with a list[list] or list[set], I think it's best to keep the adj_list struct as a dictionary of lists or sets because it allows the function to have nodes be related unique keys.

verticapy/stats/tools.py
zacandcheese: This is because there isn't a directed path between these two points. There would be an undirected edge between X<->Y and this function is only looking for directed connections.

verticapy/stats/tools.py
zacandcheese: To answer both as above. The reason why I chose to use pandas was that it was faster: it loaded the contingency_table into memory where I could do a lot of local calls instead of a bunch of system calls to Vertica. I originally had something below but with a very large dataset, constantly calling the **where** clause to create the `mini_table` was slower.

 

    unique_z = [["'{}'".format(x) for x in vdf[col].distinct()] for col in z]
    total_chi2 = 0
    total_dof = 0
    for condition in itertools.product(*unique_z):
        #SELECT a, d, c, b, count(*) AS "cnt" FROM square WHERE d=1 AND a=0 GROUP BY a , d, c , b
        sql_where_condition = " AND ".join(["{0}={1}".format(*x) for x in zip(z, condition)])
        mini_table_sql = "SELECT {0}, {1} FROM {3} WHERE {2};".format(x,y,sql_where_condition, vdf.__genSQL__())
        mini_table = vDataFrame(sql=mini_table_sql)
        pivot_table = mini_table.pivot_table_chi2(x)
        total_chi2 += pivot_table['chi2'][0]
        if pivot_table['chi2'][0] > 0:
            total_dof += pivot_table['dof'][0]
    return chi2.sf(total_chi2, total_dof) < alpha


I figured I could speed this up by first creating a contingency table in Vertica and then calling the **where** clause on that and then creating a pivot table.  A rough idea is what I have below. Except I was unable, over the summer, to figure out how to convert my **where** clause results into a proper pivot table.

    table_sql = "SELECT {0}, {1}, {2}, count(*) as nij FROM {3} GROUP BY {0}, {1}, {2}".format(', '.join(z), x, y, vdf.__genSQL__())
    contingency_table = vDataFrame(sql=table_sql)
    unique_z = [["'{}'".format(x) for x in vdf[col].distinct()] for col in z]
    
    total_chi2 = 0
    total_dof = 0
    for condition in itertools.product(*unique_z):
        #SELECT a, d, c, b, count(*) AS "cnt" FROM square WHERE d=1 AND a=0 GROUP BY a , d, c , b
        sql_where_condition = " AND ".join(["{0}={1}".format(*x) for x in zip(z, condition)])
        mini_table_sql = "SELECT {0}, {1}, nij FROM {3} WHERE {2};".format(x,y,sql_where_condition, contingency_table.__genSQL__())
        mini_table = vDataFrame(sql=mini_table_sql)
        display(mini_table.pivot_table([x,y]))
        pivot_table = mini_table.pivot_table_chi2(x)
        total_chi2 += pivot_table['chi2'][0]
        if pivot_table['chi2'][0] > 0:
            total_dof += pivot_table['dof'][0]
    return chi2.sf(total_chi2, total_dof) < alpha

So why doesn't this work?
1. `pivot_table_chi2()` expects all the data not a contingency_table
2. As shown in the files below, occasionally the pivot table (**expected result**) will have zeros so accounting for those locations given the `mini_table` proved to be difficult. Also below, I tried to run `pivot_table([x,y])` and I got those results.

Solutions:
1. Creating a quick temp table that is a contingency table so we don't need to rerun the same query to generate it because I believe currently to generate `mini_table` it runs `mini_table_sql` which reruns `contingency_table.__genSQL__()` which is time-consuming.
2. find a way to convert `mini_table` into **expected_result** within Vertica / VSQL commands. This would be almost `pivot_table` if it didn't aggregate data.

I think making this completely within Vertica is doable, but I don't think I have enough time to do it during this school semester.

files below
---

**mini_table**
image

**mini_table.pivot_table([x,y])**
image

**expected result**
image

verticapy/learn/causal_model.py
zacandcheese: I wanted to have two copies in case a user wanted to change the discretization parameters, that way we didn't lose the original relationship. So `self.input_relation` is the main file and the function operates on `self.vdf` which can be modified by the discretization function and the method itself.

verticapy/learn/causal_model.py
zacandcheese: This mainly applies to categorical columns, because numerical columns can be discretized. If there are too many categories then running this function is exponentially slower. Because of the way the function is set up, it needs to calculate the chi2 score for each state of `z` and if `z` is four columns each having 100 categories that is 100,000,000 iterations for just one conditional independence test. So it can be more, but we should warn against it for performance purposes.

verticapy/learn/causal_model.py
zacandcheese: Oops yes its suppose to be `X -> Z -> Y'

verticapy/learn/causal_model.py
zacandcheese: The intuition behind this rule can be found [here](http://mcb111.org/w06/KollerFriedman.pdf#page=125) but it is basically a stronger statement: if there is no edge between X and Y then we are sure we can direct Z -> Y.

verticapy/learn/causal_model.py
zacandcheese: If no edges got removed this iteration then the next iteration no edges will also get removed so the function can stop early and we can identify no edges got removed when `remaining_edges = sum([len(pdag[x]) for x in columns])` because `remaining_edges` is the count before this iteration and `sum([len(pdag[x]) for x in columns])` is the count after this iteration.

verticapy/learn/causal_model.py
oualib: @zacandcheese please correct it in your code

verticapy/learn/causal_model.py
oualib: @zacandcheese  please correct it in your code

verticapy/learn/causal_model.py
oualib: @zacandcheese  please add the warning in your code

verticapy/learn/causal_model.py
oualib: @zacandcheese can you please add a comment in your code

verticapy/stats/tools.py
oualib: @zacandcheese can you please create an issue on that?

verticapy/stats/tools.py
zacandcheese: This assumes they are categorical, if they are numeric the user must convert them into appropriate categories. They are conditional tests for numeric values but chi2 doesn't work.

oualib:
@zacandcheese you still have errors in your unit tests, can you please have a look?

ansleis (approved):
Made some edits to the comments.

oualib:
This PR needs an entire review. We need to adapt the code to the new architecture. @zacandcheese if you want we can work together on this.

Issue #354, opened 2022-08-15
Small Changes on the vDataFrame interaction


oualib:
I think we can do the following changes:
 - keep the index white
 - The footer is a little bit confusing, it says: "Showing 1 to 25 of 100 entries" whereas the vDataFrame has more entries. Maybe, we can write something more precise?

roypaulin:
I think it refers to max_rows which is set to 100 by default and represents the number of rows that will be displayed. Interactive tables are built on top of vdataframes so they will inherit any display option from them.

oualib:
I got it, please change the message as most of the users will not get it.

PR #370, opened 2022-10-09
Update vDataFrame interactiveness

roypaulin:
This updates the interactive mode by:
- keeping the index white
- The footer was confusing so I made a simple message that just tells the range of rows that are currently being displayed.

Here is an example of a table:
image

Closes: #354


roypaulin:
If you want a specific message for the footer, let me know.

PR #373, opened 2022-10-13
Labels

oualib:
Adding labels for each query.

TODO:
 - add label for CREATE and COMMIT queries.
 - add the identifier at the beginning of each method
 - update the contribute file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ with open("README.md", "r") as fh:
 
 setuptools.setup(
     name="verticapy",
-    version="0.10.1",
+    version="0.11.0",
     author="Badr Ouali",
     author_email="badr.ouali@vertica.com",
     url="https://github.com/vertica/VerticaPy",
@@ -33,7 +33,7 @@ setuptools.setup(
     packages=setuptools.find_packages(),
     python_requires=">=3.6",
     install_requires=[
-        "matplotlib>=2.0",
+        "matplotlib<=3.5.3",
         "numpy>=1.11.0",
         "pandas>=0.23.0",
         "python-highcharts>=0.4.1",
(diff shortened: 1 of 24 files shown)


setup.py
```diff
@@ -33,7 +33,7 @@
     packages=setuptools.find_packages(),
     python_requires=">=3.6",
     install_requires=[
-        "matplotlib>=2.0",
+        "matplotlib<=3.5.3",
```
afard, line 36: I think you do this to make the tests pass on Python 8. IMO, it is fine as temporary solution, but we should improve our tests sooner or later to make them resistant to the internal changes of matplotlib.

verticapy/tests/vDataFrame/test_vDF_utilities.py
```diff
@@ -19,7 +19,7 @@
 from math import ceil, floor
 
 # Other Modules
-import pandas, geopandas
+# import pandas, geopandas
```
afard, line 22: I wonder what is the reason of relocating the imports?

afard:
I'm curious what is your goal of labeling all the queries?

afard:
There is a library issue on Python 3.6 now. Can you please take a look?

setup.py
oualib: I agree

verticapy/tests/vDataFrame/test_vDF_utilities.py
oualib: It was because of the Python 3.6 bug

oualib:
@afard we are labelling the queries to be able to run statistics on the query_profile table. It will help us to understand the usage of each function.

afard (approved):

PR #375, opened 2022-10-14
Labels (save to query profile) + new params for groupby

oualib:
- New function save_to_query_profile
 - reindent
 - saving all important functions in the query profile table
 - rollup and having parameters for groupby

@afard can you please do the entire review
@ansleis can you please review the CONTRIBUTING.md file, the groupby function of the vdataframe and the 'save_query_profile' function in the utilities file?


verticapy/__init__.py
```diff
@@ -82,6 +85,7 @@
     "interactive": False,
     "count_on": False,
     "footer_on": True,
+    "identifier": random.randint(0, 1000000000),
```
afard, line 88: Are the identifiers supposed to be unique?
If yes, you'd better to use Python's UUID module (https://docs.python.org/3.8/library/uuid.html) or UUID generator on Vertica server (https://www.vertica.com/docs/12.0.x/HTML/Content/Authoring/SQLReferenceManual/Functions/UUID/UUID_GENERATE.htm).

verticapy/connect.py
```diff
@@ -392,7 +392,14 @@ def read_dsn(section: str, dsn: str = ""):
...
 
         options = confparser.items(section)
-        conn_info = {"port": 5433, "user": "dbadmin", "session_label": "verticapy"}
+        conn_info = {
+            "port": 5433,
+            "user": "dbadmin",
+            "session_label": "verticapy-"
+            + verticapy.__version__
+            + "-"
+            + str(verticapy.options["identifier"]),
```
afard, line 401: Vertica generates a unique SESSION_ID each time a new connection is made. IMHO, that is the most meaningful ID if you need an identifier to log the activities related to a specific VerticaPy connection.

verticapy/connect.py
```diff
@@ -512,6 +519,9 @@ def verticalab_connection():
         "password": "",
         "database": "demo",
         "backup_server_node": ["localhost"],
-        "session_label": "verticapy",
+        "session_label": "verticapy-"
+        + verticapy.__version__
+        + "-"
+        + str(verticapy.options["identifier"]),
```
afard, line 525: Ditto!

verticapy/__init__.py
oualib: Good idea, I will use it.

verticapy/connect.py
oualib: Session ID is unique but sometimes the user uses multiple queries with different sessions. We need an idea to merge them together. A verticapy session is different than a Vertica session.

verticapy/connect.py
oualib: Btw verticapython client is doing the same.

CONTRIBUTING.md
```diff
@@ -420,6 +420,16 @@ executeSQL("SELECT * FROM {0} LIMIT 2".format(titanic.__genSQL__()), method="fet
   135,
   'Montreal, PQ / Chesterville, ON']]
 ```
+The last thing to know before writing an entire function is the save_to_query_profile() method. It is used to save the method call in a specific Vertica table. This table can be used to run statistics or in order to debug some code.
+
+The function is quite simple and it uses 4 main parameters:
+ - name: it is the name of the function.
+ - path: it is the location of the function. For example, the method corr of the vDataFrame is located in vdataframe.vDataFrame.
+ - json_dict: it is the dictionary of the different parameters.
+ - query_label: it is the name to give to the identifier column in the query profile table.
```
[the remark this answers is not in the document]
oualib: Yes it can be an arbitrary name but you are right, we can give more explanations.

verticapy/datasets.py
```diff
@@ -283,6 +301,15 @@ def load_dataset(
     """
     General Function to ingest a dataset
     """
+    # Saving information to the query profile table
+    save_to_query_profile(
+        name="load_" + dataset_name,
+        path="datasets",
+        json_dict={"schema": schema, "name": name,},
```
afard, line 308: What about "copy_cols"?

CONTRIBUTING.md
```diff
@@ -420,6 +420,16 @@ executeSQL("SELECT * FROM {0} LIMIT 2".format(titanic.__genSQL__()), method="fet
   135,
   'Montreal, PQ / Chesterville, ON']]
 ```
+The last thing to know before writing an entire function is the save_to_query_profile() method. It is used to save the method call in a specific Vertica table. This table can be used to run statistics or in order to debug some code.
+
+The function is quite simple and it uses 4 main parameters:
+ - name: it is the name of the function.
+ - path: it is the location of the function. For example, the method corr of the vDataFrame is located in vdataframe.vDataFrame.
+ - json_dict: it is the dictionary of the different parameters.
```
afard, line 428: It is a guide for the potential future developers, so it is important to be clear. Are all the parameters supposed to appear hear? Or is it up the developer?

verticapy/connect.py
afard: I don't understand your claim: "A verticapy session is different than a Vertica session."? What do you mean by verticapy session?

verticapy/__init__.py
oualib: Done

verticapy/connect.py
oualib: When a user import verticapy. He can uses multiple connection to Vertica for the same import. We need to track that.

verticapy/datasets.py
oualib: we don't really need them. It is the main function used by all the load_datasets sub functions. We need to track the usage of load_titanic... and not the main one.

verticapy/connect.py
afard: Do you mean in their `_generate_session_label()` function?
https://github.com/vertica/vertica-python/blob/16ce5b020b8b65460be48f8f7e12d0fb55d79669/vertica_python/vertica/connection.py#L246

Perhaps, it is not clear to me what you want to achieve from this "session_lable". Btw, I think it would be nice if you make a similar function for generating session labels instead of duplicating your logic.

afard: Ah, you may not need a function because you have the logic only in verticapy.options.

verticapy/connect.py
afard: Does each VerticaPy connection correspond to a VerticaPy session, or the group of all connections following the same `import verticapy`?

CONTRIBUTING.md
oualib: Corrected

verticapy/__init__.py
```diff
@@ -82,6 +85,7 @@
     "interactive": False,
     "count_on": False,
     "footer_on": True,
+    "identifier": str(uuid.uuid1()).replace("-", ""),
```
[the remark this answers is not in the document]
oualib: easier to extract as we are separating verticapy - version - uuid

verticapy/hchart.py
```diff
@@ -65,13 +65,20 @@
...
 )
 from verticapy.highchart import hchartSQL
 
 # ---#
 @needs_local_scope
 def hchart(line, cell="", local_ns=None):
 
+    # Saving information to the query profile table
+    save_to_query_profile(
+        name="hchart_magic", path="hchart", json_dict={"line": line, "cell": cell,},
```
afard, line 78: Is `local_ns` intentionally excluded from json_dict?

oualib: yes because it will be too big to store everything inside

verticapy/udf.py
```diff
@@ -104,11 +115,7 @@ def import_lib_udf(udf_list: list, library_name: str, include_dependencies: list
         return True
     except Exception as e:
         os.remove(directory + "/" + file_name)
-        if conn:
-            conn.close()
         raise e
-    if conn:
-        conn.close()
```
afard, line 111: Don't we need to close the connection anymore?

ansleis (approved):
Reviewed groupby, save_to_query_profile, and contributing.md.

afard (approved):

PR #403, opened 2022-11-10
Improving read_csv, read_json and read_file

oualib:
- New parameters: flex_name and genSQL for pcsv
- New parameters genSQL and materialize for read_csv and read_json
- New parameters insert, ingest_local for read_file
- Correction of some bugs

TODO:
 unitests for read_csv, read_json, read_file, to_csv, to_json

### DIFF
--- a/verticapy/utilities.py
+++ b/verticapy/utilities.py
@@ -1447,6 +1447,7 @@ Returns
 vDataFrame
     The vDataFrame of the relation.
     """
+    version(condition=[11, 1, 1])
     from verticapy import vDataFrame
 
     # Saving information to the query profile table
@@ -3244,10 +3245,12 @@ The tablesample attributes are the same than the parameters.
             elif isinstance(val, datetime.timezone):
                 val = f"'{val}'::timestamptz"
             elif isinstance(val, (np.ndarray, list)):
+                version(condition=[10, 0, 0])
                 val = "ARRAY[{0}]".format(
                     ",".join([str(get_correct_format_and_cast(k)) for k in val])
                 )
             elif isinstance(val, dict):
+                version(condition=[11, 0, 0])
                 all_elems = [
                     "{0} AS {1}".format(get_correct_format_and_cast(val[k]), k)
                     for k in val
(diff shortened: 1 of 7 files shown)


verticapy/utilities.py
```diff
@@ -2041,20 +2162,21 @@ def read_json(
...
+            query, title="Looking if the relation exists.", method="fetchall"
+        )
+    if not (genSQL) and (column_name != []) and not (insert):
         raise NameError(
             "The table {} already exists !".format(
                 format_schema_table(schema, table_name)
             )
         )
-    elif (column_name == []) and (insert):
+    elif not (genSQL) and (column_name == []) and (insert) and not (genSQL):
```
mail4umar, line 2179: not (genSQL) is two times

mail4umar (approved):

verticapy/utilities.py
oualib: Done;

tconsigny (approved):

verticapy/vdataframe.py
```diff
@@ -11567,11 +11581,24 @@ def to_csv(
...
             current_nb_rows_written += limit
             file_id += 1
-            file.close()
-        return self
+            if n_files == 1 and path:
+                file = open(path, "w+")
+                file.write(csv_file)
+                file.close()
+            elif path:
+                file = open("{0}/{1}.csv".format(path, file_id), "w+")
```
mail4umar, line 11592: There is already "" outside of {0}. And path is also a string. It may lead to ""/folder/subfolder"/".  If it works then fine.

Otherwise:

file = open(path+"/{1}.csv".format(path, file_id), "w+")

verticapy/vdataframe.py
```diff
@@ -11887,16 +11923,31 @@ def to_json(
...
-            file.write("]")
-            file.close()
-        return self
+            json_file += "]"
+            if n_files == 1 and path:
+                file = open(path, "w+")
+                file.write(json_file)
+                file.close()
+            elif path:
+                file = open("{0}/{1}.json".format(path, file_id), "w+")
```
mail4umar, line 11939: The same as above. 
There is already "" outside of {0}. And path is also a string. It may lead to ""/folder/subfolder"/".

oualib: Not true, path is not modified and '"' are Python quotes and do not belong to the string.

PR #406, opened 2022-11-10
Adding complex datatype parameter for read_json

### LINKED ISSUE #397: read_json -> parameter use_complex_datatype
Allowing the read_json function to use complex datatype. It will call the 'read_file' function if the parameter is set to True.

mail4umar:
@oualib Please check it out. 

Currently, we cannot search for number of files in the server so wildcard search for number of files ONLY works for local. But currently infer DDL does not work for local so counting max files is essentially useless. but I have put it anyway for future. 

close #397


oualib (approved):
Approved, but you should use black to indent your code. Also, you should create a unit test for your code.

PR #409, opened 2022-11-11
Jaro Winkler + vdf.join

oualib:
- Formatting the code
 - Jaro + Jaro Winkler distances
 - vdf.join: improvement new operators

### DIFF
--- a/verticapy/vdataframe.py
+++ b/verticapy/vdataframe.py
@@ -8004,6 +8004,8 @@ vColumns : vColumn
         on_list = []
         if isinstance(on, dict):
             on_list += [(key, on[key], "=") for key in on]
+        else:
+            on_list += [elem for elem in on]
         on_list += [(key, on[key], "linterpolate") for key in on_interpolate]
         # Checks
         self.are_namecols_in([elem[0] for elem in on])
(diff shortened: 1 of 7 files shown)


oualib:
@tconsigny can you please look at this PR, it should be one of the last ones before v0.11.0 (and the last one which needs doc review)

verticapy/vdataframe.py
```diff
@@ -7896,8 +7896,27 @@ def join(
...
+                     '<' : key1  < key2
+                     '>' : key1  > key2
+                    '<=' : key1 <= key2
+                    '>=' : key1 >= key2
+                 'llike' : key1 LIKE '%' || key2 || '%'
+                 'rlike' : key2 LIKE '%' || key1 || '%'
+           'linterpolate': key1 INTERPOLATE key2
+           'rinterpolate': key2 INTERPOLATE key1
+        Some operators need 5-tuples: (key1, key2, operator, operator2, x) where
+        operator2 is one simple operator (=, >, <, <=, >=), x a float or an integer
```
mail4umar: x IS a float

verticapy/vdataframe.py
```diff
@@ -7896,8 +7896,27 @@ def join(
     ----------
     input_relation: str/vDataFrame
         Relation to use to do the merging.
-    on: dict, optional
-        Dictionary of all different keys. The dict must be similar to the following:
+    on: dict / list, optional
+        List of 3-tuples. Each tuple must include (key1, key2, operator) where
```
mail4umar: maybe write: If it is a list then....
and below we can write : if it a dict then

Maybe use two indents for each to make it more clear.

tconsigny (approved):

mail4umar (approved):

verticapy/vdataframe.py
oualib: for levenstein, it is an integer

verticapy/vdataframe.py
oualib: ok

PR #425, opened 2022-11-29
Correction of multiple small issues

oualib:
- Save to Query Profile: new parameter in set_option to choose either to save all the logs or partially some of them. Also, It does not affect the "print_sql" option anymore.
 - Flex Table: better vmap guessing, flat vmap can now exclude some columns.
 - read_csv: guess automatically the delimiter whenever it is possible.

For the doc:
 - New param exclude_columns for vDataFrame.flat_vmap
 - Parameter sep changed for read_csv function.
 - New option for the set_option function.

TODO:
 - Unit test for read_csv
 - Unit test on the exclude of flat_vmap

### DIFF
--- a/verticapy/utilities.py
+++ b/verticapy/utilities.py
@@ -2791,7 +2791,7 @@ def set_option(option: str, value: Union[bool, int, str] = None):
             You can also provide a list of specific methods to store. For example: 
             if you specify ["corr", "describe"], only the logs associated with 
             those two methods are stored. 
-            If set to False, this functiality is deactivated.
+            If set to False, this functionality is deactivated.
         sql_on             : bool
             If set to True, displays all the SQL queries.
         temp_schema        : str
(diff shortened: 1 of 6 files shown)


verticapy/utilities.py
```diff
@@ -1012,8 +1012,17 @@ def isvmap(
             116,
         ):  # 116 is for long varbinary and 115 is for long varchar
             return False
-    except:
-        return False
+    except Exception as e:
+        if "'utf-8' codec can't decode byte" in str(e):
+            try:
+                sql = f"SELECT MAPVERSION({column}) AS isvmap FROM {expr} WHERE {column} IS NOT NULL LIMIT 1;"
```
afard, line 1018: This query is similar to the one in line 1004 except that it doesn't have `, {column}` in the SELECT clause. Couldn't you just drop that from your first query? I don't understand why it is needed there.

tconsigny (approved):

afard (approved):

PR #427, opened 2022-11-29
DBLINK - PART 1

oualib:
- Creating a vDataFrame from an external relation: new parameter external
 - running an external query: new sql option external/dblink/ext
 - new function connect.set_external_connection to set an external connection


tconsigny (approved):

verticapy/sql.py
```diff
@@ -164,6 +174,18 @@ def sql(line, cell="", local_ns=None):
...
             queries = options["-c"]
 
+        # Looking at external sources
+        if "-ext" in options and options["-ext"]:
+            assert verticapy.options["connection"]["dblink"], ConnectionError(
+                "No Connection Identifier Database is defined. Use the function connect.set_external_connection to set one."
+            )
+            sql = "SELECT DBLINK(USING PARAMETERS cid='{0}', query='{1}', rowset={2}) OVER ();"
+            queries = sql.format(
+                verticapy.options["connection"]["dblink"].replace("'", "''"),
```
mail4umar, line 184: here the user will only be entering names like cid='pgbd' etc
the replace is extra i think.

oualib: if the cid is something like ABCD'EFGH, it is not probable but it is better to treat all the cases

Issue #418, opened 2022-11-15
Allowing to merge keys in flex ingestion


oualib:
When dealing with JSON, we can have similar keys having different names. It would be nice to add an option to merge them.

PR #437, opened 2022-12-05
Small corrections + new toolbox functions.

oualib:
- Small typo corrections
 - New tool box function to help on the issue #418

### DIFF
--- a/verticapy/toolbox.py
+++ b/verticapy/toolbox.py
@@ -1610,3 +1610,300 @@ class str_sql:
 
     def category(self):
         return self.category_
+
+#
+#
+# Tools to merge similar names/categories together.
+#
+# ---#
+def erase_prefix_in_name(name: str, prefix: list = []):
+    """
+-------------------------------------------------------------------------------
+Excludes the input lists of prefixes from the input name and returns it.
+When there is a match, the other elements of the list are ignored.
+
+Parameters
+---------- 
+name: str
+    Input name.
+prefix: list, optional
+    List of prefixes.
+
+Returns
+-------
+name
+    The name without the prefixes.
+    """
+    name_tmp = name
+    for p in prefix:
+        n = len(p)
+        if p in name_tmp and name_tmp[:n] == p:
+            name_tmp = name_tmp[n:]
+            break
+    return name_tmp
+
+# ---#
+def erase_suffix_in_name(name: str, suffix: list = []):
+    """
+-------------------------------------------------------------------------------
+Excludes the input lists of suffixes from the input name and returns it.
+When there is a match, the other elements of the list are ignored.
+
+Parameters
+---------- 
+name: str
+    Input name.
+suffix: list, optional
+    List of suffixes.
+
+Returns
+-------
+name
+    The name without the suffixes.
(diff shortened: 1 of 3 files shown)


verticapy/toolbox.py
```diff
@@ -1610,3 +1610,300 @@ def __round__(self, x):
 
     def category(self):
         return self.category_
+
+#
+#
+# Tools to merge similar names/categories together.
+#
+# ---#
+def erase_prefix_in_name(name: str, prefix: list = []):
```
mail4umar, line 1620: Currently if we have two prefixes but the sequence inside the name does not match the sequence of prefix list then one of the prefix can remain. if that is intentional, then it is fine.

For example:
erase_prefix_in_name(name="pre1pre2waro",prefix=["pre2","pre1"])
OUPUT: 
'pre2waro'

UPDAtE: now that i re-read the details. i think you intend this to happen. SO PLEASE IGNORE IF THAT IS THE CASE

verticapy/toolbox.py
```diff
@@ -1610,3 +1610,300 @@ def __round__(self, x):
...
+-------------------------------------------------------------------------------
+Excludes the input lists of suffixes and prefixes from the input name and 
+returns it. When there is a match, the other elements of the list are ignored.
+
+Parameters
+---------- 
+name: str
+    Input name.
+suffix: list, optional
+    List of suffixes.
```
mail4umar, line 1719: @oualib  Why do we want only 1 suffix to be removed and rest to be ignored?

verticapy/toolbox.py
```diff
@@ -1610,3 +1610,300 @@ def __round__(self, x):
...
+    order: list = ["p", "s", "w"]
+):
+    """
+-------------------------------------------------------------------------------
+Excludes the input lists of suffixes, prefixes and words from the input name 
+and returns it.
+
+Parameters
+---------- 
+name1: str, optional
```
mail4umar, line 1764: i dont think mae1 name2 are optional as the function below does erase does not have the name as optional,

verticapy/toolbox.py
```diff
@@ -1610,3 +1610,300 @@ def __round__(self, x):
...
+    order: list = ["p", "s", "w"]
+):
+    """
+-------------------------------------------------------------------------------
+Excludes the input lists of suffixes, prefixes and words from the input name 
+and looks if it belongs to a specific group.
+
+Parameters
+---------- 
+name: str, optional
```
mail4umar, line 1813: i dont think they are optional

verticapy/toolbox.py
```diff
@@ -1610,3 +1610,300 @@ def __round__(self, x):
...
+    Input Name.
+group: list, optional
```
mail4umar, line 1815: not optional. used in for loop

mail4umar (changes requested):

verticapy/toolbox.py
oualib: I wanted to only test one. these functions are used to only simplify one future function.

mail4umar (approved):

PR #463, opened 2023-01-30
optimization - Part 5

oualib:
- Still improving queries
 - formatting correctly the code
 - improving imports

### DIFF
--- a/verticapy/vcolumn.py
+++ b/verticapy/vcolumn.py
@@ -161,34 +161,31 @@ Attributes
             if self.isarray():
                 vertica_version(condition=[10, 0, 0])
                 if index_start < 0:
-                    index_start_str = str(index_start) + " + APPLY_COUNT_ELEMENTS({})"
+                    index_start_str = f"{index_start} + APPLY_COUNT_ELEMENTS({{}})"
                 else:
                     index_start_str = str(index_start)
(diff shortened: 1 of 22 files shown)


verticapy/vcolumn.py
```diff
@@ -161,34 +161,35 @@ def __getitem__(self, index):
...
+                        index_stop_str = f"{index_stop} + APPLY_COUNT_ELEMENTS({{}})"
                     else:
                         index_stop_str = str(index_stop)
                 else:
                     index_stop_str = "1 + APPLY_COUNT_ELEMENTS({})"
-                elem_to_select = "{0}[{1}:{2}]".format(
-                    self.alias, index_start_str, index_stop_str
-                ).replace("{}", self.alias)
+                elem_to_select = f"{self.alias}[{index_start_str}:{index_stop_str}]"
+                elem_to_select = elem_to_select.replace("{}", self.alias)
```
mail4umar, line 175: why do we need a replace here? self.alias has already been input in previous line

mail4umar (approved):

verticapy/vcolumn.py
oualib: because, it is a sql function which can have multiple {}

PR #468, opened 2023-02-03
Typing + Code Reorganization

oualib:
The goal of this PR is to :
 - use the new Python convention (after 3.8) -- Stop type checking manually, use the new Literal DT
 - To reorganise the folders: use a config folder for global variables and options
 - Updating the header to include Opentext and to format it correctly
 - More to come


verticapy/vdataframe.py
```diff
@@ -9589,12 +9220,15 @@ def sessionize(
...
 
-    # ---#
-    @check_dtypes
     @save_verticapy_logs
-    def score(self, y_true: str, y_score: str, method: str, nbins: int = 30):
+    def score(
+        self,
+        y_true: str,
+        y_score: str,
+        method: ...,  # TO COMPLETE Literal[vp.learn.metrics.FUNCTIONS_DICTIONNARY]
```
mail4umar, line 9228: incomplete?

mail4umar (approved):

PR #470, opened 2023-02-06
Still rearranging the folder

oualib:
- building the final structure before splitting the files into multiple subcategories.


afard:
We have a "util" folder and an empty "utilities" folder. Should we get rid of one of them?

afard (approved):

PR #472, opened 2023-02-10
IMPORT / REORG : Cleaning useless imports & Reorganising the Code

oualib:


verticapy/core/vdataframe.py
```diff
@@ -1,10503 +0,0 @@
-"""
```
mail4umar, line 1: So now vdataframe is an empty file?

verticapy/datasets/datasets.py
```diff
@@ -24,15 +24,16 @@
...
 # VerticaPy Modules
 import verticapy, vertica_python
-from verticapy.utils._decorators import (
-    save_verticapy_logs,
-    check_minimum_version,
-)
-from verticapy import vDataFrame
+from verticapy._version import check_minimum_version
+from verticapy._utils._collect import save_verticapy_logs
+from verticapy.core.vdataframe import vDataFrame
```
mail4umar, line 29: at one place I saw you do:
from verticapy.core.vdataframe.vdataframe import vDataFrame
i think better to be consistent

verticapy/utils/_decorators.py
```diff
@@ -1,84 +0,0 @@
-"""
```
mail4umar, line 1: is this an empty file? or deleted file?

mail4umar (approved):
The code has been very neatly divided into subfolders. 

Though the names of most of the sub-files are self-explanatory, but in the next step we can add some one liners at the top of the code to help users understand what a particular code file is about.

verticapy/core/vdataframe.py
oualib: it was moved to another repo

verticapy/utils/_decorators.py
oualib: deleted

verticapy/datasets/datasets.py
oualib: done

PR #475, opened 2023-02-16
Simplifying the imports

oualib:


verticapy/tests/vModel/test_model_selection.py
```diff
@@ -22,8 +22,8 @@
 import matplotlib.pyplot as plt
 
 # VerticaPy
-import verticapy
-from verticapy.core.vdataframe.vdataframe import vDataFrame
+import verticapy, drop
```
mail4umar, line 25: shouldnt it be from verticapy import drop?

mail4umar (approved):

verticapy/tests/vModel/test_model_selection.py
oualib: correct

PR #477, opened 2023-02-26
PEP8: Improving memModel Organisation

oualib:
- Creating different classes for all the algorithms.
 - Following a consistent style.
 - Renaming to follow PEP.
 - Removing TSA Model. We will soon implement the real in-DB Version.
 - Removing the verticapy schema. Beta Implementations should be pickled. We will step by step make the remaining implementations much more robust.
 - Entire support of InMemoryModel.
 - Adding a Label to know the native and non-native models.
 - Improving code readability
 - Models have more comprehensive attributes.
 - Getting the classes without computing it from the table.

TODO:
 - Drop for Non-Native Models
 - Cleaning vModels Functions.
 - Limit the calls to get_attr


verticapy/tests/vModel/test_tools.py
```diff
@@ -36,8 +36,9 @@
 from verticapy.learn.decomposition import *
 from verticapy.learn.preprocessing import *
 from verticapy.learn.tsa import *
-from verticapy.learn.tools import does_model_exist, load_model, matrix_rotation
+from verticapy.learn.tools import does_model_exist, load_model
 from verticapy.sql.create import create_verticapy_schema
+import verticapy.machine_learning.memmodel.decomposition as mm
```
mail4umar, line 41: previously you imported memmodel as "mm". But here you are importing decomposition as "mm".
please check. For consistency, maybe change name to "mm_decomp" or only "decomp"?

mail4umar (approved):

PR #478, opened 2023-02-28
Code Cleaning: features_importance + attributes.

oualib:
- Renaming get_attr -> get_vertica_attributes. We will create a more consistent get_attributes for internal attr. Cleaning also the function, we need to adapt the code in the next commit.
 - cleaning features importance to make it simple and more robust.
 - adding a property attributes to get the list of internal attr.
 - Switch the name of KernelDensity to make the calls easier.
 - Cleaning the models representation.

TODO:
 - Create a function _compute_attr for non-Native models.
 - Add the property attributes to the non-native models.
 - Moving the Tree Class to the ensemble directory.
 - Model summary.
 - splitting plotting functions.

### DIFF
--- a/verticapy/machine_learning/vertica/preprocessing.py
+++ b/verticapy/machine_learning/vertica/preprocessing.py
@@ -339,6 +339,12 @@ method: str, optional
     def _model_type(self) -> Literal["Scaler"]:
         return "Scaler"
 
+    @property
+    def _attributes(
+        self,
+    ) -> Literal["min_", "max_", "median_", "mad_", "mean_", "std_"]:
+        return Literal["min_", "max_", "median_", "mad_", "mean_", "std_"]
+
     @check_minimum_version
     @save_verticapy_logs
     def __init__(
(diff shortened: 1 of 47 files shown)


verticapy/machine_learning/vertica/preprocessing.py
```diff
@@ -139,6 +139,10 @@ def _model_subcategory(self) -> Literal["PREPROCESSING"]:
     def _model_type(self) -> Literal["CountVectorizer"]:
         return "CountVectorizer"
 
+    @property
+    def _attributes(self) -> Literal["stop_words_", "vocabulary_", "n_errors_"]:
+        return ["stop_words_", "vocabulary_", "n_errors_"]
```
mail4umar, line 144: not using return Literal["..."]

mail4umar (approved):

verticapy/machine_learning/vertica/preprocessing.py
oualib: Literal is only for typing.

PR #495, opened 2023-03-13
PLOTTING: Cleaning the different functions.

oualib:
- histogram draws now only histogram.
 - bar chart and horizontal bar charts have their own functions
 - starting to support Plotly.

### DIFF
--- a/verticapy/core/vdataframe/_plotting.py
+++ b/verticapy/core/vdataframe/_plotting.py
@@ -34,6 +34,18 @@ import verticapy.plotting._plotly as vpy_plotly_plt
 
 
 class vDFPlot:
+    @staticmethod
+    def _get_plotting_lib(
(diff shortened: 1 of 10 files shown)


verticapy/core/vdataframe/_plotting.py
```diff
@@ -239,6 +252,93 @@ def animated(
...
+        if len(columns) == 1:
+            return self[columns[0]].bar(
+                method=method,
+                of=of,
+                max_cardinality=max_cardinality[0],
+                h=h[0],
+                **style_kwargs,
+            )
+        else:
+            vpy_plt, kwargs = self._get_plotting_lib(
```
mail4umar, line 324: don't you still need vpy_plt and kwargs when the number of columns is 1?

oualib: No, we call the vDataColumn method.

mail4umar (approved):

PR #505, opened 2023-03-16
Update bar.py

mail4umar:
- now the labels include the column name without the quotation marks: ""
- updated the width to improve aspect ratio

### DIFF
--- a/verticapy/plotting/_plotly/bar.py
+++ b/verticapy/plotting/_plotly/bar.py
@@ -43,7 +43,7 @@ class BarChart(PlotlyBase):
     def _init_style(self) -> None:
         self.init_trace_style = {"marker_color": self.get_colors(idx=0)}
         self.init_layout_style = {
-            "yaxis_title": "density",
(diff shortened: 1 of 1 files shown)


verticapy/plotting/_plotly/bar.py
```diff
@@ -43,7 +43,9 @@ def _compute_method(self) -> Literal["1D"]:
     def _init_style(self) -> None:
         self.init_trace_style = {"marker_color": self.get_colors(idx=0)}
         self.init_layout_style = {
-            "width": 200 * len(self.layout["labels"]),
+            "yaxis_title": "density",
```
oualib, line 46: I don't think it is always the density. The user can also use customised aggregations. Please check out all the layout params.

oualib (changes requested):

verticapy/plotting/_plotly/bar.py
mail4umar: Done. Now i replaced it with self.layout['method'].

PR #506, opened 2023-03-17
Plotly pie sunburst

mail4umar:


verticapy/plotting/_plotly/base.py
```diff
@@ -16,6 +16,40 @@
 """
 from verticapy.plotting.base import PlottingBase
 
+import numpy as np
+
 
 class PlotlyBase(PlottingBase):
-    ...
+    @staticmethod
```
oualib, line 23: will you use this method in many classes or just the nested pie?

verticapy/plotting/_plotly/__init__.py
```diff
@@ -15,5 +15,12 @@
...
-from verticapy.plotting._plotly.barh import HorizontalBarChart
\ No newline at end of file
+from verticapy.plotting._plotly.pie import PieChart, NestedPieChart
+from verticapy.plotting._plotly.barh import HorizontalBarChart
+from verticapy.plotting._plotly.base import PlotlyBase
+
+import plotly.io as pio
+import plotly.graph_objects as go
+
+pio.templates["VerticaPy"] = go.layout.Template(layout_colorway=PlotlyBase.get_colors())
```
oualib, line 25: what are pio templates?

verticapy/plotting/_plotly/base.py
mail4umar: for now, it is only nested pie.

verticapy/plotting/_plotly/__init__.py
mail4umar: these basically set the color/theme of the plotly library. This way, i do not need to input colors to all the different plotly functions independently.

oualib: Nice.

verticapy/plotting/_plotly/base.py
oualib: ok maybe you should move this function to your plotly base class as it seems like a general method.

PR #514, opened 2023-03-22
PLOTTING: Code Cleaning + Animated Bubble

oualib:
- Code cleaning: matplotlib files are entirely cleaned.
 - Animated Bubble is now supported with the new functions.
 - Cleaning the vdf _plotting and animated files.

 -> END MATPLOTLIB CLEANING

TODO:
 - integrate highcharts functions and delete the hchart vdf function


verticapy/core/vdataframe/_plotting.py
```diff
@@ -1614,149 +1805,228 @@ def barh(
         ).draw(**kwargs)
 
     @save_verticapy_logs
-    def hist(
+    def pie(
         self,
-        by: Optional[str] = None,
         method: str = "density",
```
mail4umar, line 1810: You took out method for vDataFrame so you can take it out here as well. 
If you want to keep it, then you may want to add the "PlottingMethod"

mail4umar (approved):

verticapy/core/vdataframe/_plotting.py
oualib: I think it is wrong. Nothing was erased, it is a Github wrong formatting.

PR #544, opened 2023-04-05
PLOTTING: Plotly Integration - Outliers Plot

mail4umar:
- outlier plot for plotly
- unit tests


verticapy/plotting/_plotly/outliers.py
```diff
@@ -0,0 +1,175 @@
...
+            df2 = pd.DataFrame(
+                {
+                    self.layout["columns"][0]: X2,
+                    "category": cat_X2,
+                    self.layout["columns"][1]: Y2,
+                }
+            )
+            concatenated_df = pd.concat([df, df2], ignore_index=True)
+            fig_scatter = px.scatter(
+                concatenated_df, x="age", y="fare", color="category"
```
oualib, line 113: Why do we age columns from the titanic dataset?

verticapy/plotting/_plotly/outliers.py
```diff
@@ -0,0 +1,175 @@
...
+            cat_X2 = ["outliers"] * len(X2)
+            x_axis = ["age"] * (len(X1) + len(X2))
+            df = pd.DataFrame(
+                {
+                    self.layout["columns"][0]: X1 + X2,
+                    "category": cat_X1 + cat_X2,
+                    "x_axis": x_axis,
+                }
+            )
+            fig_scatter = px.strip(
```
oualib, line 79: the scatter is the same 1D and 2D, you'll have a Matrix with the elements.

oualib (changes requested):

verticapy/plotting/_plotly/outliers.py
mail4umar: good catch! i corrected it.

verticapy/plotting/_plotly/outliers.py
oualib: If you look at the outliers plot of matplotlib (or even highchart), scatters calls do not change from one dimension to another so they should be drawn outside of your condition.

mail4umar: image

This is because my 1d plot is a bit different. It is without the contour. I thought it was more cleaner. What do you think?

oualib: good

oualib (approved):

PR #569, opened 2023-04-12
CODE CLEANING: Plotting + PEP8

oualib:
- Correcting plot functions.
 - removing ugly code [elem for elem...]
 - handling correctly exceptions

### DIFF
--- a/verticapy/core/string_sql/base.py
+++ b/verticapy/core/string_sql/base.py
@@ -100,7 +100,7 @@ class StringSQL:
         elif len(args) == 0:
             ValueError("Method 'in_' doesn't work with no parameters.")
         else:
-            x = [elem for elem in args]
+            x = copy.deepcopy(args)
         assert isinstance(x, Iterable) and not (
             isinstance(x, str)
         ), f"Method '_in' only works on iterable elements other than str. Found {x}."
@@ -114,7 +114,7 @@ class StringSQL:
         elif len(args) == 0:
             ValueError("Method '_not_in' doesn't work with no parameters.")
         else:
-            x = [elem for elem in args]
+            x = copy.deepcopy(args)
         if not (isinstance(x, Iterable)) or (isinstance(x, str)):
             raise TypeError(
                 "Method '_not_in' only works on iterable "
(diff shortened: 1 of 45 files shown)


verticapy/core/string_sql/base.py
```diff
@@ -100,7 +100,7 @@ def _in(self, *args) -> "StringSQL":
         elif len(args) == 0:
             ValueError("Method 'in_' doesn't work with no parameters.")
         else:
-            x = [elem for elem in args]
+            x = copy.deepcopy(args)
```
mail4umar, line 103: in a previous place, you used list() to replace the [elem ...]

verticapy/core/vdataframe/_corr.py
```diff
@@ -1367,8 +1369,8 @@ def acf(
             result = self._new_vdataframe(query).corr(
                 [], method=method, focus=column, show=False
             )
-            columns = [elem for elem in result.values["index"]]
-            acf = [elem for elem in result.values[column]]
+            columns = copy.deepcopy(result.values["index"])
```
mail4umar, line 1372: does this also gives a list?

verticapy/core/vdataframe/_fill.py
```diff
@@ -453,7 +456,7 @@ def fillna(
                         sql_push_ext=self._parent._vars["sql_push_ext"],
                         symbol=self._parent._vars["symbol"],
                     )
-                except:
+                except QueryError:
```
mail4umar, line 459: You are importing it as vQueryError

mail4umar (approved):

verticapy/core/vdataframe/_corr.py
oualib: yes

verticapy/core/vdataframe/_fill.py
oualib: there are two different QueryError one from VerticaPy and one from Vertica-Python.

verticapy/core/string_sql/base.py
oualib: it depends on the use-case. List will also convert dict to list.

PR #616, opened 2023-06-05
Missing docstrings

tconsigny:
I marked the functions that might be missing docstrings. Some of these might be internal (so might not be absolutely necessary to add docstrings), but thought it best to identify them just in case. If none of these functions require descriptions, we can just close this PR.


oualib:
@tconsigny can you please watch at my comments and tell me if it is okay?

oualib (changes requested):

oualib (approved):

PR #623, opened 2023-06-09
Create pylint.yml

afard:
This adds a new test on each PR checking its code quality using pylint. The test will fail if the return score of pylint becomes less than 5.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,26 @@
+name: Pylint
+
+on: 
+  pull_request:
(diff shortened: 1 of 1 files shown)


afard:
@oualib and @mail4umar, a complete pylint is very time consuming. I suggest to check only specific things. Here is the list of Pylint checks: https://pylint.readthedocs.io/en/stable/user_guide/configuration/all-options.html

What do you think?

oualib (approved):
@mail4umar can you please check what are the important tags for pylint?

PR #631, opened 2023-06-13
619 unit testing classification metrics

vikash018:
PR for classification metrics.

Following tests are skipped as functions needs to be fixed for multi-class:
- best_cutoff
- roc_auc_score
- prc_auc_score

log_loss metrics is skipped as Vertica has default log base 10, whereas Sklearn has natural log (e)


oualib (approved):
@mail4umar can you double check?

PR #648, opened 2023-06-23
633 unit testing linearregression

vikash018:
Unit test for linear-regression

### DIFF
--- a/verticapy/tests_new/machine_learning/vertica/linear_model/test_linear_regression.py
+++ b/verticapy/tests_new/machine_learning/vertica/linear_model/test_linear_regression.py
@@ -1,0 +1,572 @@
+"""
+(c)  Copyright  [2018-2023]  OpenText  or one of its
+affiliates.  Licensed  under  the   Apache  License,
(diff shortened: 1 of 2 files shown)


oualib:
@mail4umar please review all the file and approve if everything is okay. Also test it locally.

verticapy/tests_new/machine_learning/vertica/linear_model/test_linear_regression.py
```diff
@@ -0,0 +1,574 @@
...
+from verticapy.connection import current_cursor
+import numpy as np
+import sklearn.metrics as skl_metrics
+import sklearn.linear_model as skl_linear_model
+import pytest
+import statsmodels.api as sm
+from scipy import stats
+import matplotlib.pyplot as plt
+
+REL_TOLERANCE = 1e-2
```
mail4umar, line 28: Shouldn't the results be closer than just 1e-2? I can understand in certain cases the calculation method may differ but then we can tweak the tolerance for those specific cases. It will be good to know which attributes are a bit off.

mail4umar (approved):

PR #650, opened 2023-06-26
plotting visual bug fixes

mail4umar:
- Added % sign on importance plot for matplotlib
- Added Y label for lift chart and cutoff curve in matplotlib
- Removed quotation marks for plotly density plot

### DIFF
--- a/verticapy/plotting/_plotly/density.py
+++ b/verticapy/plotting/_plotly/density.py
@@ -40,7 +40,7 @@ class DensityPlot(PlotlyBase):
             "width": 700,
             "height": 500,
             "autosize": False,
-            "xaxis_title": self.layout["x_label"],
+            "xaxis_title": self.layout["x_label"][1:-1],
             "yaxis_title": self.layout["y_label"],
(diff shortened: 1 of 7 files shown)


verticapy/plotting/_plotly/density.py
```diff
@@ -40,7 +40,7 @@ def _init_style(self) -> None:
             "width": 700,
             "height": 500,
             "autosize": False,
-            "xaxis_title": self.layout["x_label"],
+            "xaxis_title": self.layout["x_label"][1:-1],
```
oualib, line 43: You should never use this kind of logic directly in the code. Imagine that the user passes a customised layout, you'll always retrieve the first and the last character.

oualib (changes requested):

verticapy/plotting/_plotly/density.py
mail4umar: Got it. Thanks for pointing out. 
Updated it.

oualib (approved):

PR #660, opened 2023-07-03
Overwrite model parameter

### LINKED ISSUE #605: Models should not be overwritten by default
Currently, the "overwrite_model" config parameter determines if a model should be overwritten whenever a new model with same name is trained. The default value of this config parameter is true. This means that users can easily make mistakes and lose their previously archived models inside Vertica DB. IMHO, this is not an appropriate behavior.

My suggestions:
(1) Let's change the default behavior such that the model overwriting would happen only if a user has explicitly asked for it.
(2) Instead of a session scope config parameter, lets make it a model specific parameter. Therefore, it must be explicitly turned on when a vModel object is declared if overwriting behavior is desired for that model name.

afard:
The main purpose of this PR is to resolve #605. Nevertheless, I made several other improvements as well.

- The `overwrite_model` parameter is moved from the session level config to a parameter of VerticaModel class
- It makes more sense to declare the `_model_category` property only in `Supervised` and `Unsupervised` classes. Then, their subclass will inherit the right one. In case of any exceptions, the property can be overriden.
- The `model_name` and `overwrite_model` are common attributes of all models so it makes sense to have them at the base class.
- The `AutoDataPrep` class needed its own drop function.
- Unfortunately, some tests didn't cleanup the database after themselves. The problem doesn't show up in the Github tests because each time a new database is created, but it is very annoying for a developer who runs the tests against the same database. I fixed those. Although there are still leftovers, but they don't cause test failures.
- I added the tests of the new  `overwrite_model` parameter only for a few models. Nevertheless, it covers all if statements that have this parameter.


verticapy/tests/vDataFrame/test_vDF_plot_plotly.py
```diff
@@ -168,35 +168,42 @@ def acf_plot_result(load_plotly, amazon_vd):
 def regression_plot_result(load_plotly, dummy_scatter_vd):
     model = LinearRegression("LR_churn")
     model.fit(dummy_scatter_vd, ["x"], "y")
-    return model.plot()
```
afard, line 171: If we use return, the function will be called and will be entirely executed each time it is needed. In other words, the model will trained over and over. We should use yield instead which keeps the objects of the fixture for the entire of its scope. Then, we can make cleaning after the yield which would be when the fixture goes out of its scope.

verticapy/machine_learning/vertica/neighbors.py
```diff
@@ -787,6 +781,8 @@ def _vertica_fit_sql(self) -> Literal["RF_REGRESSOR"]:
     def _vertica_predict_sql(self) -> Literal["PREDICT_RF_REGRESSOR"]:
         return "PREDICT_RF_REGRESSOR"
 
+    # This is an exception. Although KernelDensity is a subclass of Regressor,
+    # but it is UNSUPERVISED.
```
afard, line 785: @oualib, can you please confirm that this exception is made on purpose?

oualib: yes

oualib (approved):

PR #676, opened 2023-07-22
Model versioning

### LINKED ISSUE #371: Model Versioning
To integrate model versioning into VerticaPy, we add a new function, named "register", to the VerticaModel class. Calling this function will execute the register_model meta-function inside Vertica and registers the model. We also implement a new class in VerticaPy, named RegisteredModel, in order to work with the registered models inside the database. It may help to look at an easy example before explaining the spec of this class in details.

**An example for registering a model**
model = RandomForestClassifier(name = "my_schema.rfc1")
model.fit("public.train_data", ["pred1","pred2","pred3"], "resp")
model.register("application_name")

rm = mv.RegisteredModel("application_name")

afard:
To integrate model versioning into VerticaPy, we add a new function, named "register", to the VerticaModel class. Calling this function will execute the register_model meta-function inside Vertica and registers the model. We also implement a new class in VerticaPy, named RegisteredModel, in order to work with the registered models inside the database.

This PR closes #371


mail4umar:
Hey @afard, we have recently changed the unit tests so you may want to consider two main things:
1- All tests are now inside the "tests_new" folder. The directories follow the same structure as the original Verticpay structure. One test file for each python file.
2- All tests should be created so that they can be run in parallel i.e. using pytest -n auto --dist=loadscopee. This will help reduce the time taken for tests in the future. 

Have a look at : https://github.com/vertica/VerticaPy/blob/master/CONTRIBUTING.md#unit-tests

p.s. you can use the schema loader to avoid dropping tables/data

afard:
> Hey @afard, we have recently changed the unit tests so you may want to consider two main things: 1- All tests are now inside the "tests_new" folder. The directories follow the same structure as the original Verticpay structure. One test file for each python file. 2- All tests should be created so that they can be run in parallel i.e. using pytest -n auto --dist=loadscopee. This will help reduce the time taken for tests in the future.
> 
> Have a look at : https://github.com/vertica/VerticaPy/blob/master/CONTRIBUTING.md#unit-tests
> 
> p.s. you can use the schema loader to avoid dropping tables/data

Done!

mail4umar (approved):
nice implementation!
but perhaps next time, we can even finer tests i.e. one assert per test function.

PR #678, opened 2023-07-26
657 unit testing tree regressors

vikash018:
This PR has UT for following models:
1. RandomForestRegressor
2. RandomForestClassifier
3. DecisionTreeRegressor
4. DecisionTreeClassifier
5. XGBRegressor
6. XGBClassifier
7. Ridge
8. Lasso
9. ElasticNet
10. LinearRegression

test_base_model_methods.py - file contains comman test methods for tree/linear models
test_linear_model.py - file contains specific test methods for linear models
test_tree_model.py - file contains specific test methods UT for tree models


verticapy/tests_new/machine_learning/vertica/linear_model/test_linear_model.py
mail4umar, line 1: One quick comment:
We should try to use one test file for every .py file in the directory. That means let's try to have a test_linear_model.py file instead of the directory. You can merge the base file into the test file. 
Or if you want to use it for other tests then add it to conftest.

vikash018: Okay. I will get rid-off linear_model and tree_model folders and bring files inside it to vertica folder. 
We have common methods for linear/tree model. To avoid redundancy in test, i have created a base test file with Model class as parameter to cover all common tests method, and separate test files to cover specific test methods that are only applicable to linear and tree models.

verticapy/tests_new/machine_learning/vertica/test_base_model_methods.py
```diff
@@ -0,0 +1,1294 @@
...
+@pytest.fixture
+def model_params(model_class):
+    """
+    fixture - model parameters
+    """
+    model_params_map = {
+        "RandomForestRegressor": (
+            "n_estimators, max_features, max_leaf_nodes, sample, max_depth, min_samples_leaf, min_info_gain, nbins",
+            [
+                # (None, None, None, None, None, None, None, None),  # accuracy does not mach with default parameters
```
mail4umar, line 112: is this a copy mistake or actual error?
# (None, None, None, None, None, None, None, None),  # accuracy does not mach with default parameters

mail4umar (changes requested):

verticapy/tests_new/machine_learning/vertica/test_base_model_methods.py
vikash018: None is for default value of parameter. I have commented it as verticapy accuracy is not matching with sklearn with default param. Also, verticapy and sklearn default param values are not same for few param. so , this test does not make sense.

mail4umar (approved):

PR #688, opened 2023-08-25
RANDOM instead of SEEDED_RANDOM

afard:
Replacing the SEEDED_RANDOM function wherever possible with the RANDOM function.


.gitignore
afard, line 1: This file asks git to ignore the `__pycache__` folders at local repositories. Otherwise, these folders appear at the result of `git status` which are annoying.

PR #694, opened 2023-09-13
Vertica hotfix version

### LINKED ISSUE #674: Do we need to capture the hotfix number of Vertica version in the vertica_version function?
Do we use the last number of the Vertica version anywhere? (https://github.com/vertica/VerticaPy/blob/master/verticapy/_utils/_sql/_vertica_version.py#L122)

The "-*" part is related to the hotfix number. I wonder if we can get rid of it if it is not used anywhere.

afard:
The "-*" part of a Vertica version is related to its hotfix number. There is no hotfix number when we internally compile Vertica from source. This PR makes having hotfix number optional in VerticaPy so that it would be easier for us when we work with VerticaPy against an in-development Vertica.

This closes #674.


oualib:
The tests are failing please look why.

oualib (approved):

PR #695, opened 2023-09-14
Adding distributed seeded random

afard:
Because of legacy concerns, we have not modified the in-DB SEEDED_RANDOM function. Instead, we introduce a new function, named DISTRIBUTED_SEEDED_RANDOM, in Vertica 24.1.0. The new function does not have the distribution problem of pseudo random numbers.

Similar to the SEEDED_RANDOM function, the new function will not be publicly documented and supported because it still does not meet general expectations of a seeded random function.


oualib (changes requested):
Why not creating a general function for this SEEDED_RANDOM because the code seems to be redundant.

afard:
> Why not creating a general function for this SEEDED_RANDOM because the code seems to be redundant.

Agreed! I'll do that.

oualib (approved):

Issue #696, opened 2023-09-14
Two unit tests fail after upgrading xgboost package from 1.7.6 to 2.0.0


afard:
We suppress these two tests temporarily while we investigate the right solution.
- verticapy/tests/vModel/test_xgb_classifier.py::TestXGBC::test_to_json_multiclass
- verticapy/tests_new/machine_learning/vertica/test_tree_model.py::TestClassificationTreeModel::test_predict_proba

Eventually, we should resolve the issue with enabling these tests again.

PR #697, opened 2023-09-14
Suppressing two tests

afard:
We suppress these two tests temporarily while we investigate the right solution.
- verticapy/tests/vModel/test_xgb_classifier.py::TestXGBC::test_to_json_multiclass
- verticapy/tests_new/machine_learning/vertica/test_tree_model.py::TestClassificationTreeModel::test_predict_proba

I have created a corresponding issue (https://github.com/vertica/VerticaPy/issues/696) so that we won't forget to resolve the problem and enabling the tests again.


verticapy/tests_new/machine_learning/vertica/test_tree_model.py
vikash018, line 1: This will skip all test for predict_proba as same test is running for based on model_class parametrize. We may need to add condition only for xgb. something like.
if model_class in ["XGBRegressor", "XGBClassifier"]:
pytest.skip("This test fails after upgrading xgboost library from 1.7.6 to 2.0.0")
else:
existing condition.

afard: You're right! I didn't notice that it is a generic test.

vikash018:
looks good to me.

vikash018 (approved):

verticapy/tests_new/machine_learning/vertica/test_tree_model.py
afard: I made an if-statement inside the function because `model_class` was not undefined for an `skipif`.

oualib (approved):

PR #713, opened 2023-09-21
Plotly Line Plot Update

mail4umar:
- added multi line
- added colors parameter


verticapy/plotting/_plotly/line.py
```diff
@@ -16,6 +16,7 @@
 """
 from typing import Literal, Optional
 
+import copy
```
oualib, line 19: is copy a standard module, if yes put it at the top.

oualib (approved):

PR #715, opened 2023-09-22
Plotly Contour Plot - Color option

mail4umar:
- also fixed title option


oualib (approved):
Approved but the tests are not passing, please look why

PR #717, opened 2023-09-22
Highcharts ACF plot bugfix

mail4umar:
- now max is not 1 for viewing
- also colors can be changed

### DIFF
--- a/verticapy/plotting/_highcharts/acf.py
+++ b/verticapy/plotting/_highcharts/acf.py
@@ -96,6 +96,7 @@ class ACFPlot(HighchartsBase):
         chart, style_kwargs = self._get_chart(chart, style_kwargs=style_kwargs)
         if "colors" in style_kwargs:
             self.init_style_confidence["fillColor"] = style_kwargs["colors"]
+            style_kwargs.pop("colors")
         chart.set_dict_options(self.init_style)
         chart.set_dict_options(style_kwargs)
         if self.layout["kind"] == "bar":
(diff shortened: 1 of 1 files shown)


oualib (approved):

verticapy/plotting/_highcharts/acf.py
```diff
@@ -46,7 +46,7 @@ def _init_style(self) -> None:
                 "title": {"text": "lag"},
                 "categories": self.data["x"].tolist(),
             },
-            "yAxis": {"title": {"text": "value"}, "max": 1},
```
oualib, line 49: leave the max 1, ACF plots can not reach more than 1 and -1

oualib (changes requested):
tests are failing and changes seem to be incorrect

verticapy/plotting/_highcharts/acf.py
oualib: @mail4umar please tell me why you deleted the max?

verticapy/plotting/_highcharts/acf.py
```diff
@@ -94,6 +94,9 @@ def draw(
         """
         kind = "PACF" if self.layout["pacf"] else "ACF"
         chart, style_kwargs = self._get_chart(chart, style_kwargs=style_kwargs)
+        if "colors" in style_kwargs:
+            self.init_style_confidence["fillColor"] = style_kwargs["colors"]
```
oualib, line 98: what if style_kwargs["colors"] is not a list or the opposite?

verticapy/plotting/_highcharts/acf.py
mail4umar: In order not to cut-off the plot to display. But since >1 is not expected so i will revert.

verticapy/plotting/_highcharts/acf.py
mail4umar: have updated

oualib (approved):

PR #731, opened 2023-09-28
sphinx conversion connection directory

tconsigny:
Converts the files in the connection module. These conversions were pretty straightforward. I hard-coded the outputs for some examples because they had passwords in the output.

### DIFF
--- a/verticapy/connection/connect.py
+++ b/verticapy/connection/connect.py
@@ -69,16 +69,21 @@ def connect(section: str, dsn: Optional[str] = None) -> None:
 
     Example
     -------
+    Display all available connections:
+
     .. code-block:: python
 
-        from verticapy.connection import *
+        from verticapy.connection import available_connections
 
-        #Displays all available connections
         available_connections()
+        
+    ``['VML', 'VerticaDSN', 'VerticaDSN_test']``
 
-    ['VML', 'VerticaDSN', 'VerticaDSN_test']
+    Connect using the VerticaDSN connection:
 
     .. code-block:: python
+
+        from verticapy.connection import connect
 
         connect("VerticaDSN")
     """
@@ -113,9 +118,12 @@ def set_connection(conn: Connection) -> None:
 
     Example
     -------
+    Create a connection using the official Vertica Python client:
+
+    .. note:: You can use any connector (ODBC, JDBC, etc.) as long it has both `fetchone` and `fetchall` methods. However, note that VerticaPy works most efficiently with the native client because of its support for various complex data types and certain Vertica optimizations.
+
     .. code-block:: python
 
-        # Creating a connection
         import vertica_python
 
         conn_info = {'host': "10.211.55.14",
@@ -125,7 +133,10 @@ def set_connection(conn: Connection) -> None:
                     'database': "testdb"}
         conn = vertica_python.connect(** conn_info)
 
-        # Setting the connection
+    Set up the connector:
+
+    .. code-block:: python
+
         from verticapy.connection import set_connection
 
         set_connection(conn)
@@ -151,9 +162,14 @@ def close_connection() -> None:
 
     Example
     -------
+    Close all current connections:
+    
+    .. warning:: When you close the connection, your session will terminate and all temporary elements will be automatically dropped.
+
     .. code-block:: python
 
         from verticapy.connection import close_connection
+
         close_connection()
     """
     gb_conn = get_global_connection()
@@ -184,19 +200,27 @@ def current_connection() -> GlobalConnection:
 
     Example
     -------
+    Get the current VerticaPy connection:
+
     .. code-block:: python
 
         from verticapy.connection import current_connection
+
         conn = current_connection()
         conn
 
-    <vertica_python.vertica.connection.Connection at 0x118c1f8d0>
+    ``<vertica_python.vertica.connection.Connection at 0x118c1f8d0>``
+
+    After the connection is established, you can execute SQL
+    queries directly:
+
+    .. note:: Please refer to your connector's API reference for a comprehensive list of its functionalities.
 
     .. code-block:: python
 
         conn.cursor().execute("SELECT version();").fetchone()
 
-    ['Vertica Analytic Database v12.0.4-0']
+    ``['Vertica Analytic Database v12.0.4-0']``
     """
     gb_conn = get_global_connection()
     conn = gb_conn.get_connection()
@@ -236,19 +260,24 @@ def current_cursor() -> Cursor:
 
     Example
     -------
+    Get the current cursor:
+
     .. code-block:: python
 
         from verticapy.connection import current_cursor
+
         cur = current_cursor()
         cur
 
-    <vertica_python.vertica.cursor.Cursor at 0x11a7b4748>
+    ``<vertica_python.vertica.cursor.Cursor at 0x11a7b4748>``
+
+    Directly execute an SQL query:
 
     .. code-block:: python
 
         cur.execute("SELECT version();").fetchone()
 
-    ['Vertica Analytic Database v12.0.4-0']
+    ``['Vertica Analytic Database v12.0.4-0']``
     """
     return current_connection().cursor()
 
@@ -277,13 +306,15 @@ def vertica_connection(section: str, dsn: Optional[str] = None) -> Connection:
 
     Example
     -------
+    Create a connection using the input DSN:
+
     .. code-block:: python
 
         from verticapy.connection import vertica_connection
 
         vertica_connection("VerticaDSN")
 
-    <vertica_python.vertica.connection.Connection at 0x106526198>
+    ``<vertica_python.vertica.connection.Connection at 0x106526198>``
     """
     return vertica_python.connect(**read_dsn(section, dsn))
 
@@ -302,13 +333,17 @@ def verticapylab_connection() -> Connection:
 
     Example
     -------
+    Get the VerticaLab connection:
+
+    .. note:: VerticaLab is a Dockerized environment designed for seamlessly using VerticaPy. This function returns the connection to the Vertica instance within the lab, allowing for necessary environment customization.
+
     .. code-block:: python
 
         from verticapy.connection import verticalab_connection
 
         verticalab_connection()
 
-    <vertica_python.vertica.connection.Connection at 0x106526198>
+    ``<vertica_python.vertica.connection.Connection at 0x106526198>``
     """
     gb_conn = get_global_connection()
     conn_info = {
--- a/verticapy/connection/connect.py
+++ b/verticapy/connection/connect.py
@@ -76,7 +76,7 @@ def connect(section: str, dsn: Optional[str] = None) -> None:
         from verticapy.connection import available_connections
 
         available_connections()
-        
+
     ``['VML', 'VerticaDSN', 'VerticaDSN_test']``
 
     Connect using the VerticaDSN connection:
@@ -163,7 +163,7 @@ def close_connection() -> None:
     Example
     -------
     Close all current connections:
-    
+
     .. warning:: When you close the connection, your session will terminate and all temporary elements will be automatically dropped.
 
     .. code-block:: python
--- a/verticapy/connection/connect.py
+++ b/verticapy/connection/connect.py
@@ -66,6 +66,21 @@ def connect(section: str, dsn: Optional[str] = None) -> None:
     dsn: str, optional
         Path to the file containing the credentials.
         If empty, the  Connection File will be used.
+
+    Example
+    -------
+    .. code-block:: python
+
+        from verticapy.connection import *
+
+        #Displays all available connections
+        available_connections()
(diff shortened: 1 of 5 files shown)


verticapy/connection/connect.py
```diff
@@ -66,6 +66,21 @@ def connect(section: str, dsn: Optional[str] = None) -> None:
     dsn: str, optional
         Path to the file containing the credentials.
         If empty, the  Connection File will be used.
+
+    Example
+    -------
+    .. code-block:: python
+
+        from verticapy.connection import *
```
oualib, line 74: please do not use import *, it was old practice and it is now quite dangerous.

Use:
```
from verticapy.connection import available_connections
```
Please also write quick lines.

Ex:
1. Let's look at all the available connections.

2. We can connect using the VerticaDSN connection.

@mail4umar what is the best way to display outputs? Should we use a specific tag: code-block?? Please advice Tyler here.

verticapy/connection/connect.py
```diff
@@ -95,6 +110,25 @@ def set_connection(conn: Connection) -> None:
     ODBC connection. This should not be confused with a
     native   VerticaPy   connection  created   by   the
     'new_connection' function.
+
+    Example
+    -------
+    .. code-block:: python
```
oualib, line 116: Can you separate the code into two blocks and bring the comment outside.

1. Let's use the official Vertica Python client.
.. note:: You can use any connectors you prefer (ODBC, JDBC, etc.), as long as they have both `fetchone` and `fetchall` methods. However, it's worth noting that VerticaPy works most efficiently with the native client due to its support for various complex data types and certain Vertica optimizations.

2. We can then set up the connector.
.. warning:: This connector is used throughout the entire API, so if it's closed, you'll need to create a new one.

verticapy/connection/connect.py
```diff
@@ -114,6 +148,13 @@ def set_connection(conn: Connection) -> None:
 def close_connection() -> None:
     """
     Closes the connection to the database.
+
+    Example
+    -------
+    .. code-block:: python
+
+        from verticapy.connection import close_connection
```
oualib, line 156: add a space in between the import and the function.

Add also a line before:
1. Let's close all the current connections.
.. warning:: When you close the connection, your session will terminate, and all temporary elements will be automatically dropped.

verticapy/connection/connect.py
```diff
@@ -140,6 +181,22 @@ def current_connection() -> GlobalConnection:
     connect  using  an  auto  connection.  Otherwise,
     VerticaPy  attempts  to connect to a  VerticaLab
     Environment.
+
+    Example
+    -------
+    .. code-block:: python
+
+        from verticapy.connection import current_connection
```
oualib, line 189: please add a space between the import and the code.
Add also a line:
Let's get the current VerticaPy connection. - (Please feel free to reformulate if it is not good)
Then another one:
Once the connection is established, you can execute SQL queries directly. 
.. note:: Please refer to your connector's API reference for a comprehensive list of its functionalities.

verticapy/connection/connect.py
```diff
@@ -176,6 +233,22 @@ def current_connection() -> GlobalConnection:
 def current_cursor() -> Cursor:
     """
     Returns the current database cursor.
+
+    Example
+    -------
+    .. code-block:: python
+
+        from verticapy.connection import current_cursor
+        cur = current_cursor()
```
oualib, line 242: same as above.

verticapy/connection/connect.py
```diff
@@ -201,6 +274,16 @@ def vertica_connection(section: str, dsn: Optional[str] = None) -> Connection:
     -------
     conn
         Database connection.
+
+    Example
+    -------
+    .. code-block:: python
+
```
oualib, line 281: same as above.

verticapy/connection/read.py
```diff
@@ -63,6 +70,41 @@ def read_dsn(section: str, dsn: Optional[str] = None) -> dict:
     -------
     dict
         dictionary with all the credentials.
+
+    Example
+    -------
```
oualib, line 75: Nice example, please add lines to explain what you do (better than comments).
Please look at what your output looks like (do you have a pic to share so I understand how it renders the result?)

verticapy/connection/write.py
```diff
@@ -145,8 +204,23 @@ def new_connection(
...
+        For example: {'user': 'ENV_USER', 'password'
+        : 'ENV_PASSWORD'}
+
+    Example
+    -------
+    .. code-block:: python
+
+        from verticapy.connection import new_connection
+
+        conn_info = {"host": "10.211.55.14",
```
oualib, line 217: PEP8

verticapy/connection/write.py
```diff
@@ -145,8 +204,23 @@ def new_connection(
...
+                     "port": "5433",
+                     "database": "testdb",
+                     "password": "XxX",
+                     "user": "dbadmin"}
+
+        new_connection(conn_info, name = "VerticaDSN")
```
oualib, line 223: add one line to explain what you do. If there is no output you can write a note like:

.. note:: If no errors are raised, the new connection was successful.

verticapy/connection/utils.py
```diff
@@ -46,6 +46,13 @@ def get_connection_file() -> str:
     -------
     string
         the full path to the auto-connection file.
+
+    Example
+    -------
+    .. code-block:: python
+
+        from verticapy.connection import get_connection_file
```
oualib, line 54: new line between import and the function.

Where is the output?

oualib (changes requested):
@tconsigny amazing, just few changes to make.

@mail4umar can you please check the output rendering. I don't know what it will look like.

verticapy/connection/read.py
```diff
@@ -30,6 +30,13 @@ def available_connections() -> list[str]:
     -------
     list
         all available connections.
+
+    Example
+    -------
+    .. ipython:: python
+
```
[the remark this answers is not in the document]
tconsigny: The output shows in the rendered version

verticapy/connection/read.py
tconsigny: image

verticapy/connection/connect.py
```diff
@@ -95,6 +115,31 @@ def set_connection(conn: Connection) -> None:
...
+                    'database': "testdb"}
+        conn = vertica_python.connect(** conn_info)
+
+    Set up the connector:
+
+    .. code-block:: python
+
+        from verticapy.connection import set_connection
+
+        set_connection(conn)
```
oualib, line 142: please add the warning:

.. warning:: This connector is used throughout the entire API. If it's closed, you'll need to create a new one. This is why, in some cases, it's better to use auto-connection. Auto-connection will automatically create a new connection if the current one is closed.

verticapy/connection/utils.py
```diff
@@ -46,6 +46,16 @@ def get_connection_file() -> str:
     -------
```
mail4umar, line 46: For the get_confparser, do we need an example? @oualib

verticapy/connection/connect.py
```diff
@@ -216,6 +330,20 @@ def verticapylab_connection() -> Connection:
     -------
     conn
         Database connection.
+
+    Example
+    -------
+    Get the VerticaLab connection:
+
+    .. note:: VerticaLab is a Dockerized environment designed for seamlessly using VerticaPy. This function returns the connection to the Vertica instance within the lab, allowing for necessary environment customization.
```
mail4umar, line 338: this should be VerticaPyLab for consistency

verticapy/connection/connect.py
```diff
@@ -201,6 +303,18 @@ def vertica_connection(section: str, dsn: Optional[str] = None) -> Connection:
...
+    -------
+    Create a connection using the input DSN:
+
+    .. code-block:: python
+
+        from verticapy.connection import vertica_connection
+
+        vertica_connection("VerticaDSN")
+
+    ``<vertica_python.vertica.connection.Connection at 0x106526198>``
```
oualib, line 317: .. note:: We utilize a Data Source Name (DSN) to establish the connection, which is stored in the file specified by the global variable VERTICAPY_CONNECTION. However, if you prefer a customized file located elsewhere, you can specify the file path accordingly.

verticapy/connection/external.py
```diff
@@ -34,10 +34,22 @@ def set_external_connection(cid: str, rowset: int = 500, symbol: str = "$") -> N
...
         $$$QUERY$$$, where QUERY represents a custom
         query.
+
+    Example
+    -------
+    .. code-block:: python
+
+        import verticapy as vp
+
+        # Set up a connection with a database with the alias "pgdb"
```
oualib, line 49: PEP formatting. And write the comment outside of the function.

.. note:: When configuring an external connection, you'll need to assign a unique symbol to identify it. This symbol will subsequently allow you to extract data from the target database using the associated identifier.

mail4umar (changes requested):
@oualib 
For the file and class Global Connection, we have very small functions, do we want a docstring for those? 

E.g.: 
    def get_connection(self) -> Connection:
        return self._connection["conn"]

verticapy/connection/read.py
```diff
@@ -63,6 +73,44 @@ def read_dsn(section: str, dsn: Optional[str] = None) -> dict:
...
+    Read the DSN information from the ODBCINI environment variable:
+
+    .. code-block:: python
+
+        from verticapy.connection import read_dsn
+
+        dsn = read_dsn("VerticaDSN")
+        dsn
+
+    | ``{'database': 'testdb',``
```
oualib, line 88: PEP8 for the output.

verticapy/connection/utils.py
```diff
@@ -46,6 +46,16 @@ def get_connection_file() -> str:
...
     string
         the full path to the auto-connection file.
+
+    Example
+    -------
```
oualib, line 51: Please add a line before to say that the connection file is the one used to store all the connections. 

.. warning:: Connections are stored in plain text format. In shared environments, it's recommended to create connections manually for added security.

verticapy/connection/write.py
```diff
@@ -67,6 +90,34 @@ def delete_connection(name: str) -> bool:
...
+        new_connection({"host": "10.20.110.10",
+                        "port": "5433",
+                        "database": "vertica_eon",
+                        "password": "vertica",
+                        "user": "dbadmin"},
+                        name = "My_New_Vertica_Connection")
+
+    Display the available connections, delete the "My_New_Vertica_Connection" 
+    connection, and confirm that it no longer appears in the available
+    conenctions:
```
oualib, line 110: typo

verticapy/connection/write.py
```diff
@@ -67,6 +90,34 @@ def delete_connection(name: str) -> bool:
...
+
+    .. ipython:: python
+
+        from verticapy.connection import available_connections, delete_connection
+
+        available_connections()
```
oualib, line 116: should you print results before and after?

verticapy/connection/write.py
```diff
@@ -145,8 +211,28 @@ def new_connection(
...
+    .. code-block:: python
+
+        from verticapy.connection import new_connection
+
+        conn_info = {
+            "host": "10.211.55.14",
+            "port": "5433",
+            "database": "testdb",
+            "password": "XxX",
+            "user": "dbadmin"}
```
oualib, line 233: the '}' should be in the next line aligned with the variable.
```
conn_info = {
      "host": "10.211.55.14",
      "port": "5433",
      "database": "testdb",
      "password": "XxX",
      "user": "dbadmin"
}
```

oualib (changes requested):
almost good. still some changes to make.

verticapy/connection/utils.py
oualib: in that case, can you write it @mail4umar ?

I am writing the new example to not waste Tyler time.

Write it as comment directly.

oualib:
> @oualib For the file and class Global Connection, we have very small functions, do we want a docstring for those?
> 
> E.g.: def get_connection(self) -> Connection: return self._connection["conn"]

same. Please write directly the example if needed to help Tyler.

verticapy/connection/write.py
tconsigny: They do print in the rendered output

oualib (approved):
looks good to me, @mail4umar do you want to add something?

verticapy/connection/write.py
```diff
@@ -67,6 +90,34 @@ def delete_connection(name: str) -> bool:
     bool
         True if the connection was deleted,
         False otherwise.
+
+    Example
+    -------
+    .. ipython:: python
+        :suppress:
```
mail4umar, line 97: @tconsigny 
I think we should change this because it may not work for every environment e.g. in my verticapylab, it throws an error obviously. Instead of executing, we can use a code-block here. Then you can just write the output manually.

PR #732, opened 2023-09-28
sphinx conversion -- the rest of datasets directory

tconsigny:
Generators.py was already converted. This PR converts the rest of the datasets directory (really just loaders.py). Pretty straightforward, but I did run into an error displaying three of the datasets, seemingly because of non-standard characters. This is the error I received:

`test_file.py:docstring of test_file.load_amazon:39: CRITICAL: Problem with "raw" directive:
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc1 in position 74196: invalid start byte`

### DIFF
--- a/verticapy/datasets/loaders.py
+++ b/verticapy/datasets/loaders.py
@@ -121,6 +121,25 @@ def load_market(schema: str = "public", name: str = "market") -> vDataFrame:
     -------
     vDataFrame
         the market vDataFrame.
+
+    Example
+    -------
+    .. code-block:: python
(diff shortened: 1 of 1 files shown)


oualib (commented):
@tconsigny Is it possible to add a sentence just before the ipython cells:

When using this loader without any arguments, the dataset is loaded using the default schema (public).

@mail4umar can you double check?

oualib (approved):
please look why the tests are failing, it is quite weird...

PR #738, opened 2023-10-03
Plotly Outliers Plot - Added color customization option

mail4umar:


oualib (approved):

oualib:
Please update your PR with the latest merge. Arash corrected the pylint issue.

oualib:
why the test is not passing?

@afard can you please help here?

oualib:
Seems to be another Github bug. @afard what can we do here? Now the titanic does not want to be ingested. Maybe we should merge everything and do a PR to debug everything at once?
Most of the PRs are docs.

@vikash018 we should finalise new UT ASAP so we divide by 6 the time to run the tests.

afard:
> Seems to be another Github bug. @afard what can we do here? Now the titanic does not want to be ingested. Maybe we should merge everything and do a PR to debug everything at once? Most of the PRs are docs.
> 
> @vikash018 we should finalise new UT ASAP so we divide by 6 the time to run the tests.

These "insufficient space" failures are our space limit on Github for running concurrent tests. I think these quotas are determined by the contract between and Github. As a long-term solution, we can ask for more space. It would be nice if we can control the number of PRs that run the tests concurrently to avoid this issue so that we'd have confidence on our own code before merging to master. That said, if you think we are in rush, we can merge the PRs without tests.
The unit tests are performed once more on master after each merge. That's another opportunity to check the health of the master branch.

PR #751, opened 2023-10-17
Finalising corr's docstring

oualib:
- finalising the corr docstring file.

@mail4umar 

Please check if everything is working well. I tried to use the HTML file of datasets to represent titanic and amazon and I am not sure it will work.

Besides, should we seed the randomness for the examples? So users can reproduce the example?

Please review ASAP so I can use this methodology on new files.


mail4umar:
@oualib That was a great idea to avoid having too many tables. I have fixed black. It should be all good now.

mail4umar (approved):

PR #758, opened 2023-10-22
Sphinx machine learning/vertica/linear model

mail4umar:
Completed all linear models.

Also made a draft for first classification model. @oualib please provide your feedback on it.


oualib:
I have updated the file. I think it looks good @mail4umar double check to see if it works on your side. Also, I think some lines are too big, we should go back to the next line more often otherwise it is not readable on small screens. Also check why each time my black is failing.

oualib (approved):

oualib (approved):

PR #766, opened 2023-10-24
Sphinx Docstring Update - Bisecting KMeans

mail4umar:
@oualib Please check. Also suggest if we want to add more details like model.cluster_i_ss_ or model.children_right_ etc.


oualib:
> @oualib Please check. Also suggest if we want to add more details like model.cluster_i_ss_ or model.children_right_ etc.

I've corrected this one.

Please add a Github issue: Add all models attributes in the docstring.

We might come to that point later.

Please see why it is failing then merge the PR.

oualib (approved):

PR #772, opened 2023-10-25
sphinx docstring - ml/vertica/naive_bayes

mail4umar:
@oualib please suggest if we want to add more details on multiclass metrics.


oualib (approved):
Approved but please double check the final rendering before merging.

PR #776, opened 2023-10-25
746 tests vdataframe math and rolling

vikash018:
UT code for math and rolling functions.

Two observations:
1) log function in apply_fun - Expected log arg from vertica doc is **log (base, expression)**. However, Verticapy is considering **log (expression, base)** i.e. in reverse order.

titanic["age"].apply_fun(func = "log")

2) mad (median absolute deviation) function in analytic is using mean instead of median.
titanic.analytic(func="mad", columns="age", name="mad")

### DIFF
--- a/verticapy/tests_new/core/vdataframe/test_math.py
+++ b/verticapy/tests_new/core/vdataframe/test_math.py
@@ -1,0 +1,885 @@
+"""
+Copyright  (c)  2018-2023 Open Text  or  one  of its
+affiliates.  Licensed  under  the   Apache  License,
+Version 2.0 (the  "License"); You  may  not use this
+file except in compliance with the License.
+
+You may obtain a copy of the License at:
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless  required  by applicable  law or  agreed to in
+writing, software  distributed  under the  License is
+distributed on an  "AS IS" BASIS,  WITHOUT WARRANTIES
+OR CONDITIONS OF ANY KIND, either express or implied.
+See the  License for the specific  language governing
+permissions and limitations under the License.
+"""
+import re
+import datetime
+import pandas as pd
+import numpy as np  # pylint: disable=unused-import
+from scipy.stats import (
+    median_abs_deviation,
+    jarque_bera,
+)  # pylint: disable=unused-import
+import pytest
+from verticapy.core.tablesample.base import TableSample
+from verticapy.tests_new.core.vdataframe import REL_TOLERANCE, ABS_TOLERANCE
+from verticapy.tests_new import functions, AggregateFun
+
+
+class TestMath:
+    """
+    test class for Text functions test
+    """
+
+    @pytest.mark.parametrize(
+        "input_type, columns",
+        [
+            ("vDataFrame", "age"),
+            ("vDataFrame_column", "age"),
+            ("vDataFrame_column", ["age", "fare", "pclass", "survived"]),
+            ("vcolumn", "age"),
+            ("vcolumn", ["age", "fare", "pclass", "survived"]),
+        ],
+    )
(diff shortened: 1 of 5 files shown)


verticapy/tests_new/core/vdataframe/test_math.py
```diff
@@ -0,0 +1,885 @@
...
+from verticapy.tests_new import functions, AggregateFun
+
+
+class TestMath:
+    """
+    test class for Text functions test
+    """
+
+    @pytest.mark.parametrize(
+        "input_type, columns",
```
mail4umar, line 37: I think we should have a really simple test for this. Titanic is probably not the most ideal dataset as it does not have negative values. 

You can create a very simple dataset like:
vp.vDataFrame({"col1":[-1, 1, 2], "col2":[-10, -2.0, 3]})

Also we do not need to try a lot of combinations as they will add time to our code test. SImply one vDataFrame with two columns and one vcolumn should be enoush.

verticapy/tests_new/core/vdataframe/test_math.py
```diff
@@ -0,0 +1,885 @@
...
+            ("age", None, "floor", "np.floor(x)"),
+            ("album_cost", "sample_data", "len", "np.size(x)"),
+            ("album_cost", "sample_data", "length", "np.size(x)"),
+            ("age", None, "ln", "np.log(x)"),
+            # (
+            #     "fare",
+            #     None,
+            #     "log",
+            #     "np.log2(x, where=x != 0)",
+            # ),  # Looks like bug in code. Expected log arg from vertica doc is log(base, expression). However, Verticapy is considering log(expression, base) i.e. in reverse order.
```
mail4umar, line 207: @oualib can comment here

verticapy/tests_new/core/vdataframe/test_math.py
```diff
@@ -0,0 +1,885 @@
...
+            )
+        else:
+            vpy_res = titanic_vd_fun["name"].get_len().sum()
+            py_res = titanic_pdf["name"].apply(len).sum()
+
+        print(f"VerticaPy Result: {vpy_res} \nPython Result :{py_res}\n")
+
+        assert vpy_res == pytest.approx(py_res)
+
+    @pytest.mark.parametrize("column, n", ([("age", 4), ("fare", 2)]))
```
mail4umar, line 320: age in titanic is itneger and not float so it may not be the ideal column.

again i would suggest just using a small dummy dataset. 

vdf= vp.vDataFrame({"col1":[-1, 1.22, 2], "col2":[-10.33, -2.0, 3]})

Maybe create one dataset for this entire exercise but keep it very simple. And broad such that it covers everything

mail4umar (changes requested):

verticapy/tests_new/core/vdataframe/test_math.py
vikash018: I have used age and fare columns. Both are having float values in dataset.
for consistency, I am using titanic dataset as its being used by other test functions in this file.

Agree, we can use small dataset for this test.

verticapy/tests_new/core/vdataframe/test_math.py
vikash018: I am using normalize function on numeric columns (i.e. focus_columns) to scale data.
example -
titanic_vdf = titanic_vd_fun[focus_columns].normalize().fillna(0)
1        -1.9502503129565278  
2      -0.010561423955012732  
3       -0.35693443984814044  
4         0.6129100046526171  
5         2.8296973063686344 

Test combinations are basic scenarios (one column, multiple columns and no columns for vDataFrame). it's based on input of abs function for vDataFrame and vcolumn.

oualib (approved):

oualib:
Creating another PR to solve the two issues.

PR #787, opened 2023-10-30
supporting new versions of matplotlib

oualib:
test, let see if UT are passing


oualib:
@mail4umar when you're done with the current docstring. I want you to work on this PR to support the new matplotlib version. Many things changed.

PR #788, opened 2023-10-30
Updating version and support

oualib:


oualib:
@afard it works! We support Py 3.12 after few changes!

mail4umar (approved):

PR #797, opened 2023-10-31
Correcting linear models hierarchy

oualib:
- correct: get_attributes


verticapy/machine_learning/vertica/linear_model.py
VerticaDataScientist, line 1: It's good, you need to correct the hierarchy of Lasso and Ridge also

VerticaDataScientist (changes requested):

VerticaDataScientist (approved):
All good!

PR #822, opened 2023-11-02
PoissonRegressor UT

vikash018:
PoissonRegressor UT


oualib (approved):

oualib:
Excellent: do you check for linear model if we get the same coefficients?

vikash018:
Yes. Our coefficients are matching with sklearn. Only we need to set very low value of alpha (Constant that multiplies the L2 penalty term).
PoissonRegressor(alpha=0.00005, tol=1e-06)

oualib:
@vikash018 is it part of linear models tests?

vikash018:
yes.

PR #827, opened 2023-11-03
Add explode array function 

VerticaDataScientist:
The function will split and expand an array-like column (a varchar array with or without commas as separators between array elements) into multiple individual columns within a Vertica DataFrame.

example:
example


oualib:
@VerticaDataScientist be careful on your PR.
1. Some hidden files are added to it: they should be removed.
2. Install and use the latest version of black to indent your code.
3. Imports should be at the top of the file (use PEP8 convention). Please try to not add any dependencies: use standard Python module.
4. Please use RST convention for your docstring: @mail4umar can help you.

Please correct those points and I'll go through the code.

mail4umar (approved):

PR #830, opened 2023-11-04
Adding TIME SERIES models (AR, MA, ARMA, ARIMA)

oualib:
Entire implementation of TS models.
TODO:
1. @vikash018 Unit Tests
2. @mail4umar Entire Docstrings

Those should be done in two different PRs


oualib:
when it will be done:
1. @vikash018 I will need you for UT and we will use scipy to compare the coefficients and score.
2. @mail4umar I will need you to write an entire example
3. @VerticaDataScientist @mat-shyR I need you to describe what you need
4. @glarik please tell me if I've missed something in my implementation

PR #843, opened 2023-11-08
autoregressor unit test code

vikash018:
autoregressor unit test code


oualib (approved):
one question why adding those tests in linear models and not in a separate file?
Otherwise, amazing work!

vikash018:
> one question why adding those tests in linear models and not in a separate file? Otherwise, amazing work!

most of the TS model functions are same as liner model, so i added it linear models.

PR #928, opened 2023-11-23
[Docstring] Cleaning & Formatting

oualib:
- Cleaning and Formatting all the docstrings.
 - More to come...

### DIFF
--- a/verticapy/sdk/vertica/udf/load.py
+++ b/verticapy/sdk/vertica/udf/load.py
@@ -39,60 +39,59 @@ def import_lib_udf(
     Parameters
     ----------
     udf_list: list
-        List of tuples that includes the different functions.
+        List of tuples that includes the different
+        functions.
 
-        **function**     :
-                            [function]  Python   Function.
-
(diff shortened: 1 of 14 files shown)


verticapy/sdk/vertica/udf/load.py
```diff
@@ -39,60 +39,59 @@ def import_lib_udf(
...
-                            Example: {"param1": int,
-                            "param2": str}
-
-        **new_name**     :
-                            [str] New   function   name  when
-                            installed in Vertica.
+        List of tuples that includes the different
+        functions.
+
+         - function: function
```
mail4umar, line 45: I am not very sure about this use.
When using " - " for list bold with ":", there should not be any text in that line after ":".

mail4umar, line 45: I am not very sure about this use.
When using " - " for list bold with ":", there should not be any text in that line after ":".

mail4umar (approved):

PR #930, opened 2023-11-27
Sphinx doc visual modifications

mail4umar:
- removed images from some tiles
- made the tiles more responsive
- added shadows
- modified the look of version dropdown


oualib (approved):
Looks good to me

PR #1061, opened 2023-12-21
Prototype of collection

jslaunwhite-microfocus:
# VER-90670

This PR contains a prototype of what collecting resources as a parquet files looks like.

Key Features
* Each qprof collection activity is called a "step", broken out into its own class
* The collector iterates over steps to produce parquet files
* The collector generates a metadata.json file
* The collector bundles the metadata.json and the parquet files into a single output file
* The process is testable: each step is testable and so is the overall output

Didn't build yet, but know that we want to build
* The collect_all function should return a URL that allows the user to download files
* The collect_all function should infer a browser temp directory when using a jupyter notebook
* Better docstrings to describe the different pieces
* Continue extending tests

Next steps
* Discuss the design with everyone, ensure that we agree on the way forward
* Integrate with qprof class

### DIFF
--- a/verticapy/performance/vertica/collection_metadata.py
+++ b/verticapy/performance/vertica/collection_metadata.py
@@ -1,0 +1,54 @@
+"""
+Copyright  (c)  2018-2024 Open Text  or  one  of its
+affiliates.  Licensed  under  the   Apache  License,
+Version 2.0 (the  "License"); You  may  not use this
+file except in compliance with the License.
+
+You may obtain a copy of the License at:
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless  required  by applicable  law or  agreed to in
+writing, software  distributed  under the  License is
+distributed on an  "AS IS" BASIS,  WITHOUT WARRANTIES
+OR CONDITIONS OF ANY KIND, either express or implied.
+See the  License for the specific  language governing
+permissions and limitations under the License.
+"""
+import datetime
+import json
+import logging
+import os
+from typing import List
+
+from .file_metadata import FileMetadata
+
+
+class CollectionMetadata:
+    def __init__(self):
+        # List of FileMetadata
+        self.files = []
+        self.date = datetime.datetime.now()
+        self.version = "1.0"
+        self.logger = logging.getLogger(type(self).__name__)
+
+    def add_file(self, file_info: dict) -> None:
+        self.files.append(file_info)
+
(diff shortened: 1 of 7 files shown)


verticapy/performance/vertica/__init__.py
```diff
@@ -15,3 +15,6 @@
 permissions and limitations under the License.
 """
 from verticapy.performance.vertica.qprof import QueryProfiler
+from verticapy.performance.vertica.collection_metadata import CollectionMetadata
+from verticapy.performance.vertica.step00_vertica_version import Step00VerticaVersion
```
oualib, line 19: I really don't like the name, will we create a file for each step? I don't think we will even implement all of them.

verticapy/performance/vertica/collection_metadata.py
```diff
@@ -0,0 +1,54 @@
+"""
```
oualib, line 1: should it be in another folder named "collection" and we can then rename this file metadata.py?

oualib (changes requested):
My comments:
1. We should create a separated folder for the collection.
2. For the information collection, we do not need to work with steps but more with copy of tables so 1 file for each system table. For the version, or other "small info" we can merge them together in one file or just create one PARQUET file for each.
3. Look why your tests are failing.

jslaunwhite-microfocus:
>     1. We should create a separated folder for the collection.

Which folder would you prefer? `performance/vertica/collection?`
 
>     2. For the information collection, we do not need to work with steps but more with copy of tables so 1 file for each system table. For the version, or other "small info" we can merge them together in one file or just create one PARQUET file for each.

I agree with part of the statement above. 
- Yes, I think we can decouple some of the collection from the analysis, and just collect a few of the tables. 
- And yes, it is better to merge some of the info into a single table.
- Do we want to support "offline" analysis? For this small data, we could support analysis from the parquet files without load into a DB. It makes things easier for users. But I'm not sure if that is something to prioritize.
- I want to break the qprof steps into pieces. There's a few ways to do that.... we should discuss.

>     3. Look why your tests are failing.

All legs of the test timed out, including test legs that didn't have a new test from me. I don't have a button to resubmit tests on the same commit. Is there somewhere I can re-run tests? I am also unable to add reviewers... perhaps I need more permissions?

oualib:
I am closing this PR as we've changed our direction for the collection.

jslaunwhite-microfocus:
Correct! This PR should be closed because we have a different direction for collection

Issue #1098, opened 2024-01-03
Go to discussions for bug reports, feature requests and support


afard:
We would like to reorganize the way our customers interact with the project for reporting a bug or requesting a new feature. It seems to be more helpful if potential-bugs and new ideas are first raised in the [Discussion panel](https://github.com/vertica/VerticaPy/discussions). Then, a corresponding issue can be created when the discussion leads to a concreate conclusion.

PR #1100, opened 2024-01-03
Reshaping the template of the Issues tab

afard:
Per a request from @oualib, I investigate how we can reshape our "Issues" tab. The changes proposed in this PR is going to provide three options when trying to create a new issue:

- Trackable Change: This will create a new issue similar to what we had before.
- Bug Reports (and other issues): This will navigate the page to the "Potential bug" category in the "Discussion" tab.
- Feature Request / Enhancement: This will navigate the page to the "Ideas" category in the "Discussion" tab.

This PR will resolve issue #1098.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Bug Reports (and other issues)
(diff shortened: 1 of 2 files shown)


afard:
After merging this PR, we should update https://github.com/vertica/VerticaPy/discussions/474.
We can also pin the issue https://github.com/vertica/VerticaPy/issues/1098, so it appears at the top of the "Issues" page as a guide.

oualib (approved):

PR #1102, opened 2024-01-05
Sphinx plot transparency

mail4umar:
Below plots were fixed to have transparent backgrounds:
- Heatmap (corr, pivot etc)
- SVM (classification curve/plot)
- Geo Plot


oualib (changes requested):
remove the gif from this PR. Besides, they are the old ones and not the most recents. Also the SVM is much nicer in the previous plot.

oualib (approved):

PR #1134, opened 2024-01-24
Update Examples - Part 1

mail4umar:
Following Jupyter Notebooks are updated:
- all inside `examples/learn` folder **except comodities**. (This is because `commodities.ipynb` uses the VAR algorithm that is not available for verticapy 1.0.1. )
- `africa_education` inside `examples/understand`
- `insurance` inside `examples/business/insurance`


examples/business/insurance/insurance.ipynb
afard, line 1: (1) Have you updated all the figures based on a newly executed experiment? Why the type of "age" in the table resulted from "data.head(5)" is displayed "Integer", but it is displayed as "Int" in the table resulted from "data.describe(method='all')"?

(2) The histogram chart of age is not displayed.

(3) Have you intentionally added "help(data.hist)"?

(4) Isn't supposed to be a bar chart after each "data.bar(...)"?

(5) Why there is nothing resulted from "data.pivot_table(['age','sex'])"?

(6) No chart displayed after "data["charges"].hist(method = "count")"!

(7) Nothing displayed after "data.corr(focus = 'charges')"

(8) Nothing displayed after "result.plot(ts = 'id', ...)"

(9) Nothing displayed after "rf_model.features_importance()"

(10) No chart after "stepwise"! Is the new behavior?

examples/learn/iris/iris.ipynb
afard, line 1: There is no chart displayed!

afard:
@mail4umar, I looked at only two modified notebooks so far. It seems that they both have lost their lost their charts. I suspect it would be a common issue on all of the modified notebooks. Therefore, I'll review them all once you have fixed the problem.

examples/business/insurance/insurance.ipynb
mail4umar: @afard I also cannot see the charts on github display, but if i download the notebook i can see them. probably because the plots are using Plotly library and they create html objects instead of static PNG. Can you confirm by downloading?

What do you suggest should we do? Go with the more unappealing matplotlib library instead so the plots can be seen on github view?

afard: Okay, I guess it is fine as these are the examples in the repository and not the website.

afard: Do you have any idea why the type of columns have been changed on different vDFs? For example, the column **bmi** is initially **Numeric(10)**, but then it is changed to **Numeric(8,4)** in the data exploration section.

examples/learn/iris/iris.ipynb
afard, line 1: "We can **we** a clear linear separation" -> "We can **see** a clear linear separation"

afard, line 1: I wonder what is "0%|          | 0/3 [00:00<?, ?it/s]" displayed after some cells!

examples/understand/africa_education/africa_education.ipynb
afard, line 1: You have added a redundant "import verticapy as vp" before data-loading. There is already one before printing the version.

afard (approved):
I have added a few non-critical comments. It would be nice if you can resolve them before your merge.

examples/business/insurance/insurance.ipynb
mail4umar: Not really sure. But it is worth inspecting for future.

examples/learn/iris/iris.ipynb
mail4umar: it is just the progress of execution.

PR #1156, opened 2024-02-12
Unpack a collection bundle

jslaunwhite-microfocus:
## VER-91630

Logic for validating and unpacking a collection bundle.

Added pyarrow optional dependency for loading data with pandas.

Steps in this PR:
1. Untar the file
2. Identify the version of file contents
3. Match tables to files
4. Read the files with pandas

In the next PR we will load the tables into the database. It's reasonable to load the tables from a client connection because the tables are relatively small: a few MB of compressed data.


verticapy/performance/vertica/collection/collection_tables.py
```diff
@@ -139,8 +153,37 @@ def getAllCollectionTables(
...
     """
+    if version == BundleVersion.V1:
+        return _getAllTables_v1(target_schema, key)
+
+    raise ValueError(f"Unrecognized bundle version {version}")
+
+
+ALL_TABLES_V1 = [
+    AllTableTypes.COLLECTION_EVENTS,
+    AllTableTypes.COLLECTION_EVENTS,
```
afard, line 164: Have you repeated this element on purpose? Is it the reason you have created the ALL_TABLES_V1 list instead of iterating over AllTableTypes?
A line of clarifying in-code comment would be nice.

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -54,33 +59,97 @@ def __init__(
...
+        The DDL statements are:
+           CREATE TABLE ... IF NOT EXISTS
+           CREATE PROJECTION ... IF NOT EXISTS
+        """
+        return self._skip_create_table
+
+    @skip_create_table.setter
+    def skip_create_table(self, val: bool) -> None:
+        if not isinstance(val, bool):
+            raise ValueError(
```
afard, line 91: I think it is better to raise `TypeError` here instead of `ValueError`.

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -54,33 +59,97 @@ def __init__(
...
+        ProfileImport will raise and exception if a bundle lacks
+        any of the expected files and raise_when_missing_files is True.
+        Otherwise, it will write a warning to the log. Default value is False.
+        """
+        return self._raise_when_missing_files
+
+    @raise_when_missing_files.setter
+    def raise_when_missing_files(self, val: bool) -> None:
+        if not isinstance(val, bool):
+            raise ValueError(
```
afard, line 108: Ditto!

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -54,33 +59,97 @@ def __init__(
...
+                f"Cannot set raise_when_missing_files to value of type {type(val)}. Must be type bool."
+            )
+        self._raise_when_missing_files = val
+
+    @property
+    def tmp_path(self) -> Path:
+        """
+        ProfileImport will raise and exception if a bundle lacks
+        any of the expected files and raise_when_missing_files is True.
+        Otherwise, it will write a warning to the log. Default value is False.
```
afard, line 118: Apparently, you made a copy-paste, but then forgot to edit the comments.

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -54,33 +59,97 @@ def __init__(
...
+        """
+        return self._tmp_path
+
+    @tmp_path.setter
+    def tmp_path(self, val: os.PathLike) -> None:
+        if not isinstance(val, str) and not isinstance(val, os.PathLike):
+            raise ValueError(
```
afard, line 125: Ditto!

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -54,33 +59,97 @@ def __init__(
...
         if not os.path.exists(self.filename):
             raise FileNotFoundError(f"File {self.filename} does not exist")
 
-        # Recall that open will raise errors if
-        # the file does not exist
-        self.input_file_obj = open(self.filename, "r")
+        unpack_dir = self._unpack_bundle()
+        self.bundle_version = self._calculate_bundle_version(unpack_dir)
+        self._check_for_missing_files(unpack_dir, self.bundle_version)
+        self._load_vdataframes(unpack_dir, self.bundle_version)
```
afard, line 140: Doesn't make sense to load all VDFs as part of  check_file? It sounds like a time consuming operation more than a check.

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -101,7 +170,7 @@ def _schema_exists_or_raise(self) -> None:
     def _tables_exist_or_raise(self) -> None:
         tables_in_schema = self._get_set_of_tables_in_schema()
         all_tables = getAllCollectionTables(
-            target_schema=self.target_schema, key=self.key
+            target_schema=self.target_schema, key=self.key, version=self.bundle_version
```
afard, line 173: The `getAllCollectionTables` function will throw an error when `self.bundle_version` is None. To avoid it, wouldn't be better to set it to `BundleVersion.LATEST`, like what is done in the `check_schema`?
What about changing `version` to  `BundleVersion.LATEST` when it is None inside `getAllCollectionTables`?

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -141,11 +210,92 @@ def _create_schema_if_not_exists(self) -> None:
 
     def _create_tables_if_not_exists(self) -> None:
         all_tables = getAllCollectionTables(
-            target_schema=self.target_schema, key=self.key
+            target_schema=self.target_schema, key=self.key, version=self.bundle_version
```
afard, line 213: The same concern here when `self.bundle_version` is None.

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -141,11 +210,92 @@ def _create_schema_if_not_exists(self) -> None:
...
+
+        Returns
+        --------
+        Returns a path
+        """
+
+        self.tarfile_obj = tarfile.open(self.filename, "r")
+        names = "\n".join(self.tarfile_obj.getnames())
+        print(f"Files in the archive: {names}")
+        # There are other ways to generate a
```
afard, line 234: Is the comment complete?

verticapy/performance/vertica/collection/profile_import.py
afard: There are several other similar cases as well. Is it a desired behavior for these function to fail if they are called prior to the `check_file` function?

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -54,33 +59,97 @@ def __init__(
...
         Checks to see that the file exists and that we can open it
         """
         if not os.path.exists(self.filename):
             raise FileNotFoundError(f"File {self.filename} does not exist")
 
-        # Recall that open will raise errors if
-        # the file does not exist
-        self.input_file_obj = open(self.filename, "r")
+        unpack_dir = self._unpack_bundle()
+        self.bundle_version = self._calculate_bundle_version(unpack_dir)
```
afard, line 138: I wonder if it makes sense to move this line to the `__init__`. It seems that several functions rely on the value of `self.bundle_version`.

verticapy/tests_new/performance/vertica/collection/test_import_simple.py
```diff
@@ -31,6 +32,16 @@ class TestProfileImport:
     Collection of tests for ProfileImport class
     """
 
+    @pytest.fixture
```
afard, line 35: The default scope is "function". Is it good enough?

setup.py
```diff
@@ -54,6 +54,7 @@
             "descartes>=1.0.0",
             "geopandas>=0.8.0",
             "shapely>=1.6.0",
+            "pyarrow>=14.0.0"
```
mail4umar, line 57: You should add a sentence in the PR detail about adding a new requirement, along with some context of why is it needed. Just for traceability.

verticapy/tests_new/performance/vertica/collection/test_import_simple.py
```diff
@@ -55,14 +66,42 @@ def test_no_such_schema(self):
...
+        pi = ProfileImport(
+            target_schema="schema_not_used", key="no_such_key", filename=fname
+        )
+        pi.skip_create_table = True
+        with pytest.raises(FileNotFoundError, match=f"File {fname} does not exist"):
+            pi.check_file()
+
+    def test_untar_file(self, tmp_path_with_test_bundles):
+        """
+        Confirm failure when user specifies a file that does not exist
```
mail4umar, line 90: i think the docstring is not accurate. it is not confirming failure.

mail4umar (changes requested):
Maybe in the next PR you can also add in docstring for the class ProfileImport and others which follow the docstring syntax including parameters, attributes, examples etc.

verticapy/performance/vertica/collection/collection_tables.py
jslaunwhite-microfocus: I'll add a comment in the code to clarify.

The reason is: we know that we will change the version of the bundle, and that future versions will have more tables. So we need to version the list of tables that we expect in a bundle. It's overkill now: there is only one version. But I expect the bundle versions to change soon when we collect more files.

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: +1 TypeError is better, I will change this.

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: Will update this comment

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: For testing purposes, I want to be able to call `check_schema` and `check_file` separately. 

I will refactor with a centralized way to set version if it has not been set.

setup.py
jslaunwhite-microfocus: I'll add a sentence in the PR detail

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: oops, this is incomplete, thanks for catching that. I'll add the text that is missing

verticapy/tests_new/performance/vertica/collection/test_import_simple.py
jslaunwhite-microfocus: Hrm... I made this function scope because `tmp_dir` is also function scope. But I suppose this has more work associated with it. I will change it to be class scope.

verticapy/tests_new/performance/vertica/collection/test_import_simple.py
jslaunwhite-microfocus: Oops, copy/paste error, I will fix this

verticapy/performance/vertica/collection/collection_tables.py
afard: It makes sense for creating the `ALL_TABLES_V1` list. What about repeating `AllTableTypes.COLLECTION_EVENTS` in the list?

setup.py
jslaunwhite-microfocus: Added to PR description

verticapy/performance/vertica/collection/collection_tables.py
jslaunwhite-microfocus: Good catch! There is an unwanted dupe. Fixed in d87dc12b, test added in 4d861c4b

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: 66219ca2 and added a test

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: Fixed in 66219ca2

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: Fixed in 66219ca2

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: fixed in 66219ca2

verticapy/performance/vertica/collection/profile_import.py
afard: It doesn't seem to be changed!

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: I put the initialization of `bundle_version` in the `__init__` function and removed any other attempts to initialize it elsewhere. 

The creation of `unpack_dur` needs to wait until `check_file` because it is valid to create an ProfileImport object without a file to import.

I want to avoid having a `__init__` function that has many requirements and lots of work. I'd like to be able to create instances of `ProfileImport` that only test database interaction, or only test file interaction. It allows the tests to be more precise. 

I expect to have a factory function that will handle common cases of creating a ProfileImport object.

mail4umar (approved):

setup.py
```diff
@@ -54,6 +54,7 @@
             "descartes>=1.0.0",
             "geopandas>=0.8.0",
             "shapely>=1.6.0",
+            "pyarrow>=14.0.0",
```
oualib, line 57: why this dependency?

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -16,11 +16,17 @@
...
+import random
+from typing import Set, List
+
+import pandas as pd
 
 from verticapy._utils._sql._sys import _executeSQL
+from verticapy.core.vdataframe import vDataFrame
 
-from .collection_tables import getAllCollectionTables, AllTableTypes
+from .collection_tables import getAllCollectionTables, AllTableTypes, BundleVersion
```
oualib, line 29: please use always full import path.

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -54,33 +59,97 @@ def __init__(
...
+                f"Cannot set skip_create_table to value of type {type(val)}. Must be type bool."
+            )
+        self._skip_create_table = val
+
+    @property
+    def raise_when_missing_files(self) -> bool:
+        """
+        ProfileImport will raise and exception if a bundle lacks
+        any of the expected files and raise_when_missing_files is True.
+        Otherwise, it will write a warning to the log. Default value is False.
```
oualib, line 101: use "`" for any Python object. Ex: ``True`` or ``raise_when_missing_files=True``

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -141,11 +210,92 @@ def _create_schema_if_not_exists(self) -> None:
...
+
+        tmp_dir.mkdir()
+
+        self.tarfile_obj.extractall(tmp_dir)
+
+        print(f"Extracted files: {[x for x in tmp_dir.iterdir()]}")
+
+        return tmp_dir
+
+    def _calculate_bundle_version(self, unpack_dir: Path) -> BundleVersion:
```
oualib, line 249: A quick description is needed to explain what you do.

verticapy/performance/vertica/collection/profile_import.py
```diff
@@ -141,11 +210,92 @@ def _create_schema_if_not_exists(self) -> None:
...
+        metadata_file = unpack_dir / "profile_metadata.json"
+        if not metadata_file.exists():
+            self.logger.info(f"Did not find metadata file {metadata_file}")
+            return BundleVersion.V1
+
+        return BundleVersion.LATEST
+
+    def _check_for_missing_files(
```
oualib, line 257: same

oualib (changes requested):

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: Refactored `_load_vdataframes` so that it is done as part of a new function, `load_file`

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: I have set the bundle version to be defaulted to `BundleVersion.LATEST` in `ProfileImport`.

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: I have refactored this so that `self.bundle_version` is `LATEST` by default, instead of unset. That is sufficient for the kinds of tests we want to run.

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: Instead of fixing the comment, I did what the comment would have suggested, which is to make a directory unique by using a timestamp.

verticapy/tests_new/performance/vertica/collection/test_import_simple.py
jslaunwhite-microfocus: I needed to change the fixture to use `tmp_path_factory` at class scope. But it worked! Now we will copy files fewer times.

verticapy/tests_new/performance/vertica/collection/test_import_simple.py
jslaunwhite-microfocus: Fixed this

setup.py
jslaunwhite-microfocus: It is required to by `pandas` to load parquet files. I had it installed locally already when I installed pandas. But the CI did not have it installed and complained that it could not read parquet files without it.

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: b15bbf23 should have the change

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: fixed

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: fixed

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: added a docstring comment

verticapy/performance/vertica/collection/profile_import.py
jslaunwhite-microfocus: added a docstring comment

jslaunwhite-microfocus (commented):
@oualib  - I've resolved all the changes you requested

jslaunwhite-microfocus:
> @oualib - I've resolved all the changes you requested

All tests passed and all changes made. Looking for another review to ensure this is ready to go.

oualib (approved):

jslaunwhite-microfocus:
Thank you for resolving and merging this branch! Thanks to everyone for helpful review comments.

PR #1165, opened 2024-02-23
Added code to overwrite the tree when Nan values are present

mail4umar:
Previously we would get ``nan`` values for training_row_count. That produced an error. 

Now such ``nan`` values from the server side will be treated as follows:

if traning_row_count== NaN:
    if not node.isRoot():
        set traning_row_count to 0
    else:
        discard the tree


oualib (approved):

glarik (approved):
Looks good to me!

PR #1168, opened 2024-02-26
[QPROF] Correcting Bugs

oualib:
- Correcting multiple bugs.
 - new parameter: display the max size of storage access
 - Managing Zero Div error + NoneType


verticapy/performance/vertica/qprof_utility.py
```diff
@@ -113,6 +113,32 @@ def _get_metrics() -> list:
...
+    def _get_metrics_name(metric: str, inv: bool = False) -> str:
+        look_up_table = {
+            "cost": "Query plan cost",
+            "rows": "Estimated row count",
+            "exec_time_ms": "Execution time in ms",
+            "est_rows": "Estimated row count",
+            "proc_rows": "Processed row count",
+            "prod_rows": "Produced row count",
+            "rle_prod_rows": "Produced RLE row count",
+            "clock_time_us": "Clock time in us",
```
mail4umar, line 126: maybe change it to microseconds? instead of us?
or use "\u00B5"

verticapy/tests_new/performance/vertica/test_qprof.py
```diff
@@ -930,21 +930,20 @@ def test_get_qplan(self, qprof_data, return_report, print_plan):
...
-            # "mem_all_mb",  # ZeroDivisionError
-            # ["cost", "proc_rows"],
+            ("cost", "Query plan cost"),
+            ("rows", "Estimated row count"),
+            ("exec_time_ms", "Execution time in ms"),
+            ("est_rows", "Estimated row count"),
+            ("proc_rows", "Processed row count"),
+            ("prod_rows", "Produced row count"),
+            ("rle_prod_rows", "Produced RLE row count"),
+            ("clock_time_us", "Clock time in us"),
```
mail4umar, line 942: same microseconds

mail4umar (changes requested):

verticapy/tests_new/performance/vertica/test_qprof.py
oualib: don't think it is necessary otherwise we will have to do it for everything

verticapy/performance/vertica/qprof_utility.py
oualib: don't think it is necessary otherwise we will have to do it for everything

PR #1173, opened 2024-02-27
Qprof Interface Update + format_query

mail4umar:
made the following changes to QueryProfielrInterface.get_qplan_tree;

- Added theQuery Execution Time for each query
- Added a display of the Query 

Also fixed format_query:

Previously there was no way to fetch the html representation of a formatted query if you don't print it. I updated it such that this can now give the html without printing:

```
format_query(query, print_sql=False)
```

### DIFF
--- a/verticapy/_utils/_sql/_format.py
+++ b/verticapy/_utils/_sql/_format.py
@@ -699,13 +699,13 @@ def format_query(
         display(Markdown(html_res))
     if indent_sql:
         res = indent_vpy_sql(res)
-    elif print_sql:
+    if print_sql:
         print(res)
-    if display_success:
-        return res, html_res
-    elif only_html:
+    if only_html:
         return html_res
-    return res
+    elif display_success:
+        return res, html_res
+    return res, None
 
 
 """
--- a/verticapy/_utils/_sql/_format.py
+++ b/verticapy/_utils/_sql/_format.py
@@ -585,7 +585,10 @@ def _format_keys(
 
 
 def format_query(
-    query: SQLExpression, indent_sql: bool = True, print_sql: bool = True
+    query: SQLExpression,
+    indent_sql: bool = True,
+    print_sql: bool = True,
+    only_html: bool = False,
 ) -> SQLExpression:
     """
     Query Formatter. It is used
@@ -698,7 +701,11 @@ def format_query(
         res = indent_vpy_sql(res)
     elif print_sql:
         print(res)
-    return res, html_res
+    if display_success:
+        return res, html_res
+    elif only_html:
+        return html_res
+    return res
 
 
 """
--- a/verticapy/_utils/_sql/_format.py
+++ b/verticapy/_utils/_sql/_format.py
@@ -639,28 +639,25 @@ def format_query(
     """
     display_success = print_sql and conf.get_import_success("IPython")
     res = clean_query(query)
-    if display_success:
-        html_res = res
-    else:
-        html_res = None
+    html_res = res
+
     # STRINGS
-    if display_success:
-        html_res = re.sub(
-            r"(\"(.)+\")",
-            COL_TAG_L + r" \1 " + COL_TAG_R,
-            html_res,
-        )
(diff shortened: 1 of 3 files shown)


roypaulin:
image

This is what it looks like. I don't like this approach because:
  - If the query is long, it won't fit in that space
  - If in the feature we add more settings, there will be even less space for execution time + query
  - The style(width, margin, padding, alignment etc,...) is not consistent
I think the settings box should only contain settings info and not display anything.
I have a few ideas about how we can do this and I can share them with you when you have time.

oualib:
> image
> 
> This is what it looks like. I don't like this approach because:
> 
> * If the query is long, it won't fit in that space
> * If in the feature we add more settings, there will be even less space for execution time + query
> * The style(width, margin, padding, alignment etc,...) is not consistent
>   I think the settings box should only contain settings info and not display anything.
>   I have a few ideas about how we can do this and I can share them with you when you have time.

I think it is a good solution at the moment, and you can then improve it later?

roypaulin:
> > image
> > This is what it looks like. I don't like this approach because:
> > 
> > * If the query is long, it won't fit in that space
> > * If in the feature we add more settings, there will be even less space for execution time + query
> > * The style(width, margin, padding, alignment etc,...) is not consistent
> >   I think the settings box should only contain settings info and not display anything.
> >   I have a few ideas about how we can do this and I can share them with you when you have time.
> 
> I think it is a good solution at the moment, and you can then improve it later?

I am not asking to do everything on this ticket but to at least make something that would be easy to improve. If later there is not enough space and we have to move those widgets, it will take more effort.
But if right now displaying the execution time + query is the AC then yes this is good enough

oualib (changes requested):
Your changes on the format files make the overall tests bugging... Those functions are really sensitive...

mail4umar:
> > > image
> > > This is what it looks like. I don't like this approach because:
> > > 
> > > * If the query is long, it won't fit in that space
> > > * If in the feature we add more settings, there will be even less space for execution time + query
> > > * The style(width, margin, padding, alignment etc,...) is not consistent
> > >   I think the settings box should only contain settings info and not display anything.
> > >   I have a few ideas about how we can do this and I can share them with you when you have time.
> > 
> > 
> > I think it is a good solution at the moment, and you can then improve it later?
> 
> I am not asking to do everything on this ticket but to at least make something that would be easy to improve. If later there is not enough space and we have to move those widgets, it will take more effort. But if right now displaying the execution time + query is the AC then yes this is good enough

@roypaulin , i added an overflow so that when the query is longer, we get a scroll bar. We can control the size of query/overflow conveniently.

roypaulin:
> > > > image
> > > > This is what it looks like. I don't like this approach because:
> > > > 
> > > > * If the query is long, it won't fit in that space
> > > > * If in the feature we add more settings, there will be even less space for execution time + query
> > > > * The style(width, margin, padding, alignment etc,...) is not consistent
> > > >   I think the settings box should only contain settings info and not display anything.
> > > >   I have a few ideas about how we can do this and I can share them with you when you have time.
> > > 
> > > 
> > > I think it is a good solution at the moment, and you can then improve it later?
> > 
> > 
> > I am not asking to do everything on this ticket but to at least make something that would be easy to improve. If later there is not enough space and we have to move those widgets, it will take more effort. But if right now displaying the execution time + query is the AC then yes this is good enough
> 
> @roypaulin , i added an overflow so that when the query is longer, we get a scroll bar. We can control the size of query/overflow conveniently.

Can you show me an image with a long query where overflow is applied?

roypaulin:
The query header is too big.

mail4umar:
> The query header is too big.

I agree. We can decrease that considering the utilization of the space.

mail4umar:
> Your changes on the format files make the overall tests bugging... Those functions are really sensitive...

I updated it such that the old functionality remains and another one is added.

roypaulin (approved):

oualib (approved):

PR #1198, opened 2024-03-26
Update look + fix bug

mail4umar:
- Fix bug for no schema
- Update formatting for display


oualib (approved):
@mail4umar does it close any issue?

PR #1209, opened 2024-04-29
Update qprof interface

mail4umar:
- Changed name of query id numbers from "trasaction_idx" to "query_idx"
- Updated the metric selecting widget to a dropdown which was suggested by Sruthi

### DIFF
--- a/verticapy/performance/vertica/qprof_interface.py
+++ b/verticapy/performance/vertica/qprof_interface.py
@@ -118,13 +118,13 @@ class QueryProfilerInterface(QueryProfiler):
 
         # graph headers
         self.cpu_header = widgets.HTML(
-            value=f"<h1><b>CPU Time by node and path_id - [transaction_idx: {self.index_widget.value}]</b></h1>"
+            value=f"<h1><b>CPU Time by node and path_id - [query_idx: {self.index_widget.value}]</b></h1>"
         )
         self.qpt_header = widgets.HTML(
-            value=f"<h1><b>Query Plan Tree - [transaction_idx: {self.index_widget.value}]</b></h1>"
+            value=f"<h1><b>Query Plan Tree - [query_idx: {self.index_widget.value}]</b></h1>"
         )
         self.qsteps_header = widgets.HTML(
-            value=f"<h1><b>Query Execution Steps - [transaction_idx: {self.index_widget.value}]</b></h1>"
+            value=f"<h1><b>Query Execution Steps - [query_idx: {self.index_widget.value}]</b></h1>"
         )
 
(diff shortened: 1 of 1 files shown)


verticapy/performance/vertica/qprof_interface.py
```diff
@@ -137,13 +137,22 @@ def get_qplan_tree(self, use_javascript=True):
...
-        tags = widgets.TagsInput(
-            value=["Execution time in ms", "Produced row count"],
-            allowed_tags=[
-                QprofUtility._get_metrics_name(i) for i in QprofUtility._get_metrics()
-            ],
-            allow_duplicates=False,
+        options_dropwdown = [
+            QprofUtility._get_metrics_name(i) for i in QprofUtility._get_metrics()
+        ]
+        dropdown1 = widgets.Dropdown(
```
roypaulin, line 143: A few questions about the dropdown choice:
 - Why do you think it is better than a single `TagsInput`?
 - Any plan of selecting more than 2 metrics at a time? If yes then dropdowns will not scale well.
 - If we end up using dropdowns for metrics, we should prevent setting the same metric in both dropdowns

roypaulin:
Can you also post a screenshot here? I would like to see how it looks with 2 dropdowns.

verticapy/performance/vertica/qprof_interface.py
mail4umar: Because of user feedback. They thought it was not obvious/intuitive how to change the values. And the dropdown is more intuitive. 
For now, we do not plan to do that. If we did, it would be straightforward to add another dropdown. 
I think we should not restrict user from same selection because it could create a bad user experience e.g.

User wants to see row count (dropdown # 1) and time in ms (dropdown # 2). But then they decide to swap it. Now if they want to make dropdown # 1 = time in ms, it will not allow. That is an interim step they need to perform in order to switch it.

mail4umar:
> Can you also post a screenshot here? I would like to see how it looks with 2 dropdowns.

image

verticapy/performance/vertica/qprof_interface.py
roypaulin: > User wants to see row count (dropdown # 1) and time in ms (dropdown # 2). But then they decide to swap it. Now if they want to make dropdown # 1 = time in ms, it will not allow. That is an interim step they need to perform in order to switch it.
Got you. But they should also not be able to set the same value on both dropdowns. ex: dropdown # 1 = time in ms and dropdown # 2 = time in ms. How are you going to handle this case?

roypaulin (approved):

PR #1216, opened 2024-05-21
Creating hidden functions for QueryProfiler class

mail4umar:
There are certain functions that get overwritten by QueryProfielrInterface.

I have created hidden functions which can be called separately.

The report output has also been updated to reflect this.


oualib (changes requested):
No need for an entire doc for those hidden function. You can write a brief description of why you create them. It will make the PR more readable.

oualib (approved):

PR #1266, opened 2024-08-21
[QPROF] Supporting NULL Value + More Checks

oualib:
- Checks on operator_id.
 - Supporting NULL values.

Should close:
 - https://jira.verticacorp.com/jira/browse/VER-96161?filter=-1
 - https://jira.verticacorp.com/jira/browse/VER-96160?filter=-1

### DIFF
--- a/verticapy/performance/vertica/qprof.py
+++ b/verticapy/performance/vertica/qprof.py
@@ -165,7 +165,7 @@ class QueryProfiler:
             tables using many parameters. It will
             make the process much more expensive.
     ignore_operators_check: bool, optional
-        If set to ``True`` additional tests are done
+        If set to ``False`` additional tests are done
         on ``operator_id``s.
     iterchecks: bool, optional
         If set to ``True``, the checks are done
--- a/verticapy/performance/vertica/qprof.py
+++ b/verticapy/performance/vertica/qprof.py
@@ -1612,7 +1612,6 @@ class QueryProfiler:
                 + " FROM "
                 + " FULL JOIN ".join(jointables)
             )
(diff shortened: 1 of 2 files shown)


verticapy/performance/vertica/qprof.py
```diff
@@ -164,6 +164,9 @@ class QueryProfiler:
             This parameter will aggregate on many
             tables using many parameters. It will
             make the process much more expensive.
+    ignore_operators_check: bool, optional
+        If set to ``True`` additional tests are done
+        on ``operator_id``s.
```
mail4umar, line 169:
```suggestion
        If set to ``False`` additional tests are done
        on ``operator_id``s.
```

mail4umar (changes requested):

verticapy/performance/vertica/qprof.py
```diff
@@ -1572,6 +1577,7 @@ def _check_v_table(self, iterchecks: bool = True) -> None:
                                 transaction_id = {tr_id}
                             AND statement_id = {st_id}
                             LIMIT 1"""
+                        query = self._replace_schema_in_query(query)
```
mail4umar, line 1580: @oualib There is an error. I think this line is causing the schema name and key id to be duplicated in the query.

mail4umar (changes requested):

verticapy/performance/vertica/qprof.py
oualib: Please remove them, there are two.

mail4umar (approved):

PR #1287, opened 2024-09-26
[QPROF] Many enhancements

oualib:
Possibility to filter based on tooltip.
Should close:
 - https://jira.verticacorp.com/jira/browse/VER-97128?filter=-1
 - https://jira.verticacorp.com/jira/browse/VER-97129?filter=-1


mail4umar (approved):
Works as required!

One thing from previous "op_filter". Please add a check so that it does not accept anything else other than a list. 

For example:

```
qprof1.get_qplan_tree(op_filter = "Join")
```

Should error out instead of returning a tree with no path_id highlighted.

PR #1296, opened 2024-10-02
Verbosity

oualib:
- Implementing VerticaPy verbosity.

Should close:
 - https://jira.verticacorp.com/jira/browse/VER-93167?filter=-1


verticapy/_utils/_sql/_format.py
```diff
@@ -695,15 +691,13 @@ def format_query(
             .replace("\n", "<br>")
             .replace("    ", "&nbsp;&nbsp;&nbsp;&nbsp;")
         )
-    if display_success:
```
mail4umar, line 698: You are changing this print. Is that intentional? 

Even below where you change to 'elif print_sql'. You are adding the RETURN to print_sql. Just double check if that was intentional.

verticapy/core/vdataframe/_aggregate.py
```diff
@@ -1022,13 +1022,12 @@ def describe(
...
                                 col_to_compute += [column]
                                 break
-                    elif conf.get_option("print_info"):
-                        warning_message = (
-                            f"The vDataColumn {column} is not numerical, it was ignored."
-                            "\nTo get statistical information about all different "
-                            "variables, please use the parameter method = 'categorical'."
-                        )
-                        warnings.warn(warning_message, Warning)
+                    warning_message = (
```
mail4umar, line 1025: the elif should be switched to else

mail4umar (approved):

verticapy/core/vdataframe/_aggregate.py
oualib: thanks

PR #1344, opened 2025-02-21
Three tier bar chart for get_qplan()

mail4umar:
working towards: https://ot-internal.saas.microfocus.com/ui/entity-navigation?p=4001/37001&entityType=work_item&id=4245291

Issue: In some cases the execution steps could have three-tier e.g. 

PreparePlan:LocalPlan:ParquetdataQueue

Previous the two-tier drill would not correctly show this.

Now with this implementation, it would correctly.

### DIFF
--- a/verticapy/performance/vertica/qprof.py
+++ b/verticapy/performance/vertica/qprof.py
@@ -2685,25 +2685,39 @@ class QueryProfiler:
         div = self._get_interval_str(unit)
         query = f"""
             SELECT
-                REPLACE(COALESCE(REGEXP_SUBSTR(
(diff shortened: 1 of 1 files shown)


afard:
Can you please take care of the failing test?

verticapy/tests_new/performance/vertica/test_qprof.py
```diff
@@ -703,6 +703,7 @@ def test_get_queries(self, qprof_data):
             "h",
         ],
     )
+    @pytest.mark.skip("This test is replaced by other specific tests blow it")
```
afard, line 706: I'm curious why you don't remove the test if it is redundant?

afard (approved):

PR #1346, opened 2025-03-17
BugFix - QueryProfiler - Added handling for namespace

mail4umar:
Previously, we never encountered namespace issue for QueryProfiler. 

But it turns out that Eon mode has namespaces. The current function `schema_relation` which is used inside the `get_qplan_tree` would give an error if a relation name like "name_space.schema.table" was provided. Now, I have added a handling for such a case.

The function `schema_relation` would return below for the case of "name_space.schema.table":

schema = 'name_space.table'
relation_name = 'table'

### DIFF
--- a/verticapy/_utils/_sql/_format.py
+++ b/verticapy/_utils/_sql/_format.py
@@ -1663,6 +1663,11 @@ def schema_relation(relation: Any, do_quote: bool = True) -> tuple[str, str]:
             )
             if rel_transf.split('"')[2] != ".":
                 raise ParsingError("The format of the input relation is incorrect.")
+        elif quote_nb == 0 and dot_nb == 2:
+            # Handle namespace.schema.table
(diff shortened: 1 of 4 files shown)


verticapy/_utils/_sql/_format.py
```diff
@@ -1663,6 +1663,11 @@ def schema_relation(relation: Any, do_quote: bool = True) -> tuple[str, str]:
             )
             if rel_transf.split('"')[2] != ".":
                 raise ParsingError("The format of the input relation is incorrect.")
+        elif quote_nb == 0 and dot_nb == 2:
```
afard, line 1666: Can you please add a unit test for this case?

mail4umar: We previously did not have any unit tests. I have added unit tests for all cases.

afard (approved):

PR #1347, opened 2025-04-07
Bugfix - Query Tree - Updating the JS libraries

mail4umar:
Updated library versions for index.js

Suddenly the query tree stopped displaying. Upon investigation I found that the WASM link was giving a 404 error. Now it works after this change.


roypaulin:
Do you have a screenshot?

mail4umar:
> Do you have a screenshot?

Yes, here you go. And i have confirmed that zoom works.

image

roypaulin (approved):

PR #1350, opened 2025-06-17
Update for QueryProfiler UI - Added Queue time, and Run Time

mail4umar:
Previous "Execution Time" in the Summary tab was a bit misleading. 
Now we are adding three times:

- Completion Time (total time since requested by user)
- Queue Time (time in queue)
- Run Time (the time actually taken by query to run including planning)

### DIFF
--- a/verticapy/performance/vertica/qprof.py
+++ b/verticapy/performance/vertica/qprof.py
@@ -1883,6 +1883,7 @@ class QueryProfiler:
         self.start_timestamp = []
         self.end_timestamp = []
         self.query_successes = []
+        self.queue_times = []
         # Extracting transaction_ids and statement_ids from the list of tuples
         transaction_ids = [t[0] for t in self.transactions]
         statement_ids = [t[1] for t in self.transactions]
@@ -1906,7 +1907,8 @@ class QueryProfiler:
                 query_duration_us,
                 q2.start_time,
                 q2.end_time,
-                q2.success
+                q2.success,
+                ROUND(EXTRACT(EPOCH FROM (q3.acquisition_timestamp - q3.queue_entry_timestamp)), 3) AS queue_time_sec
             FROM 
                 v_internal.dc_requests_issued AS q0
             FULL JOIN
@@ -1915,10 +1917,13 @@ class QueryProfiler:
             FULL JOIN
                 v_monitor.query_consumption AS q2 
                 USING (transaction_id, statement_id)
+            LEFT JOIN
+                v_monitor.resource_acquisitions AS q3
+                USING (transaction_id, statement_id)
             WHERE 
                 {transaction_id_condition}
                 AND {statement_id_condition};
-                """
+        """
         query = self._replace_schema_in_query(query)
         res = _executeSQL(
             query,
@@ -1934,6 +1939,7 @@ class QueryProfiler:
                 "query_start_timestamp": row[5],
                 "query_end_timestamp": row[6],
                 "query_success": row[7],
+                "queue_time_sec": row[8],
             }
         for tr_id, st_id in self.transactions:
             if (tr_id, st_id) not in transactions_dict:
@@ -1950,9 +1956,11 @@ class QueryProfiler:
                 self.start_timestamp += [info["query_start_timestamp"]]
                 self.end_timestamp += [info["query_end_timestamp"]]
                 self.query_successes += [info["query_success"]]
+                self.queue_times += [info["queue_time_sec"]]
         self.request = self.requests[self.transactions_idx]
         self.qduration = self.qdurations[self.transactions_idx]
         self.query_success = self.query_successes[self.transactions_idx]
+        self.queue_time = self.queue_times[self.transactions_idx]
 
     def _get_current_nodes(self):
         """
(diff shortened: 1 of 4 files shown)


verticapy/tests_new/core/vdataframe/test_agg.py
```diff
@@ -462,7 +462,10 @@ def test_vdf(self, titanic_vd, func_name, vpy_func, py_func):
         print(
             f"Function name: {vpy_func} \nVerticaPy Result: {vpy_res} \nPython Result :{py_res}\n"
         )
-        assert vpy_res == pytest.approx(py_res, abs=1e-3)
+        if func_name in ["all", "any"]:
```
lcoffinOT, line 465: Could you explain this change to me? Is it related to the queue/run time display change?

verticapy/performance/vertica/qprof.py
```diff
@@ -1906,7 +1907,8 @@ def _set_request_qd(self):
                 query_duration_us,
                 q2.start_time,
                 q2.end_time,
-                q2.success
+                q2.success,
+                ROUND(EXTRACT(EPOCH FROM (q3.acquisition_timestamp - q3.queue_entry_timestamp)), 3) AS queue_time_sec
```
lcoffinOT, line 1911: I know in the example we looked at yesterday, this was 0, and you said you saw that in other tests too. Have you been able to set up a query that reliably does get queued and show something here? I wonder if `EXTRACT(EPOCH...` is not the right function

lcoffinOT (changes requested):
Might want to add a new test case validating the new timing fields. Maybe check that they're always positive values (or 0), and that queue time is less than completion time.

verticapy/tests_new/core/vdataframe/test_agg.py
mail4umar: For some unexplainable reason, this test suddenly started failing. 

When I looked closely into it, it was not properly setup. 

There are two functions all (https://www.vertica.com/python/documentation/1.1.x/html/api/verticapy.vDataFrame.all.html#verticapy.vDataFrame.all) and any (https://www.vertica.com/python/documentation/1.1.x/html/api/verticapy.vDataFrame.any.html#verticapy.vDataFrame.any).

For verticapy they result in 0/1. Whereas similar functions in our python test, it outputs False/True. 

So i added a condition to switch to Boolean for a proper comparison.

verticapy/performance/vertica/qprof.py
mail4umar: Yes, I have tested it. I compared it with the actual table

```
SELECT
  transaction_id,
  statement_id,
  pool_name,
  queue_entry_timestamp,
  acquisition_timestamp,
  ROUND(
    EXTRACT(EPOCH FROM (acquisition_timestamp - queue_entry_timestamp)),
    3
  ) AS queue_time_sec
FROM v_monitor.resource_acquisitions
WHERE acquisition_timestamp > CURRENT_TIMESTAMP - INTERVAL '5 minutes'
ORDER BY acquisition_timestamp DESC
LIMIT 5;
```

lcoffinOT (commented):
Hey @mail4umar thanks for the update! I still think we should include new test cases with this PR as well (see my previous review comment https://github.com/vertica/VerticaPy/pull/1350#pullrequestreview-2939663622 ). If this needs review again while I'm out, can you pull Danny in?

mail4umar:
> Might want to add a new test case validating the new timing fields. Maybe check that they're always positive values (or 0), and that queue time is less than completion time.

This will be a bit problematic. Because we will not create a scenario where the resources are strained and there is a queue time. 

In the code it is only pulling out the values from the resource_acq table. I am sure that table will have checks in place for making sure that there are no negative values etc.

DMickens:
> Might want to add a new test case validating the new timing fields. Maybe check that they're always positive values (or 0), and that queue time is less than completion time.

My thought is that we are pulling directly from Vertica for this data. If this test failed, it would point to an issue in Vertica, not qprof. For that reason I believe that test would be unnecessary

verticapy/performance/vertica/qprof_interface.py
```diff
@@ -112,7 +112,9 @@ def __init__(
         self._query_display_info = widgets.HTML(
             value=f"""
         <b>Query Execution Success:</b> {self._success_html if self.query_success else self._failure_html} <br>
-        <b>Execution Time:</b> {self.get_qduration()} (seconds)<br>
+        <b>Completion Time:</b> {self.get_qduration()} (seconds)<br>
+        <b>Queue Time:</b> {self.queue_time} (seconds) <br>
+        <b>Run Time:</b> {float(self.get_qduration())-float(self.queue_time)} (seconds) <br>
```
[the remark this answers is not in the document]
mail4umar: Good catch! Thanks

This is how it looks like now

image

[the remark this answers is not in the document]
lcoffinOT: @DMickens by "align the outputs" did you mean the text should be vertically aligned for those three fields?

lcoffinOT:
Looks good to me, but seems like it's failing the tests now? Or am I reading the 'Some checks were not successful' box wrong?

PR #1352, opened 2025-09-12
Bugfix - QueryProfilerInterface gave error when there was no queue time

mail4umar:
i have created helper functions to process the queue_time and run_time values for the QueryProfiler object.

For the QueryProfilerInterface, I have added checks for cases when the values could be None. 

This will make the QueryProfilerInterface support older qprofs.

### DIFF
--- a/verticapy/performance/vertica/qprof.py
+++ b/verticapy/performance/vertica/qprof.py
@@ -2533,6 +2533,135 @@ class QueryProfiler:
             return None
         return float(self.qduration / self._get_interval(unit))
 
+    # Step 2b: Query queue time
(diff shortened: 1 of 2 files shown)


mail4umar:
> Should we add a test for the case where queue time or duration is Null?

As this is the UI part, we do not have tests for UI. 
QueryProfiler object has tests, but QueryProfilerInterface does not. 

We probably need a nig effort to incorporate the tests for UI. At that time we can incorporate such cases as well. But for now issues are being caught as people interact with the UI.

lcoffinOT (approved):

PR #1365, opened 2026-03-30
Bugfix - Validate path should return the path string

mail4umar:
Fix validate_path to return the validated path, preventing NoneType errors when using VERTICAPY_CONNECTION and ensuring correct connection file handling.

This was brought up in discussion: https://github.com/vertica/VerticaPy/discussions/1359


vikash018 (approved):
looks good.

